PR #3, opened 2018-09-30
Update LogoutEvent.php

snickom:
Class name must be LogoutEvent not LoginEvent.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no


hiqsol:
Thank you!

PR #7, opened 2018-10-27
Clear asset files in error view

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/yii2/issues/16101

Currently test views for this package are taken from `yii-core/tests/data`. I believe package specific views should be moved here but since I don't know the plans for it I'll prepare separate PR for `yii-core` with missing view.

### DIFF
--- a/src/ErrorHandler.php
+++ b/src/ErrorHandler.php
@@ -259,7 +259,10 @@ class ErrorHandler extends \yii\base\ErrorHandler
             return ob_get_clean();
         }
 
-        return $this->app->getView()->renderFile($_file_, $_params_, $this);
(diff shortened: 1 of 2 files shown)


hiqsol:
Files used in this repository and not used in the core must be moved here. Together with history, which is a bit time and labour intensive. That's why I haven't finished it yet.

It would be easier to collect list of files to move and then do them altogether.

samdark:
Closing since errors are now handled differently.

PR #27, opened 2018-11-18
move View to yii\web\Controller

kids-return:
move View to yii\web\Controller

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -11,7 +11,7 @@ use yii\base\Action;
 use yii\base\InlineAction;
 use yii\helpers\Url;
 use yii\helpers\Yii;
-use yii\view\View;
+use yii\exceptions\InvalidArgumentException;
 use yii\view\ViewContextInterface;
 
 /**
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -11,16 +11,22 @@ use yii\base\Action;
 use yii\base\InlineAction;
 use yii\helpers\Url;
 use yii\helpers\Yii;
+use yii\view\View;
+use yii\view\ViewContextInterface;
 
 /**
  * Controller is the base class of web controllers.
  *
  * For more details and usage information on Controller, see the [guide article on controllers](guide:structure-controllers).
  *
+ * read-only.
+ * @property View|\yii\web\View $view The view object that can be used to render views or view files.
+ * @property string $viewPath The directory containing the view files for this controller.
+ *
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @since 2.0
  */
-class Controller extends \yii\base\Controller
+class Controller extends \yii\base\Controller implements ViewContextInterface
 {
     /**
      * @var bool whether to enable CSRF validation for the actions in this controller.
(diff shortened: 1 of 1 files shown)


kids-return:
https://github.com/yiisoft/yii-core/pull/75

hiqsol:
Why?

kids-return:
@hiqsol 
`yii-core` don't need depend `view`
`yii-console` and `yii-rest` also don't need `view`
`\yii\console\Controller`extend `\yii\base\Controller`
`\yii\console\Controller`  need implement yii\view\ViewContextInterface

kids-return:
https://github.com/yiisoft/yii-console/blob/56ce5310d1febf3884529141146e4b6fd5eeae3a/src/Controller.php#L41

kids-return:
image

kids-return:
or `yii-core` require `yiisoft/view` ?

rob006:
What was the reason for extracting separate `view` package?

samdark:
View package could be used to render templates which could potentially be used in CLI apps.

rob006:
> View package could be used to render templates which could potentially be used in CLI apps.

Same for API. Then why we extracted view stuff to separate package? It could stay in core.

samdark:
@hiqsol what do you think?

kids-return:
I think the API and CLI do not require a view in many cases.
If you need to use view, you can do this by doing:
```json
require: {
    "yiisoft/view": "^3.0@dev",
    "yiisoft/yii-base-api": "^3.0@dev"
     or
    "yiisoft/yii-base-cli": "^3.0@dev"
}
```
I think the most important thing about `yii2.0` to `yii3.0` is separation.
In the case where we encounter this always, we should first consider how to separate, instead of returning to `yii2.0`, otherwise `yii3.0` will have no meaning.
the work we are doing now will be in vain，Causing a lot of time wasted!
May wait until `php8` `php9` `php10` has been released, we are still in the development of `yii3.0`

hiqsol:
> It could stay in core

Actually, almost everything could stay in core.
The reasons to split packages are:

1. allow using the package independently
2. use other packages without the package.

In this case, both reasons are applicable. But the second reason is applicable to a greater degree.
A common API application and some of console applications don't need views rendering.
That's why no need for `view` in the core.

`yiisoft/yii-core` package is not actually dependent on `yiisoft/view` package.
It only depends on `ViewContextInterface`.
And it raises chicken (core) and egg (view) question: where this interface should come from?
That's why PSR interfaces are so important - they allow to create uncoupled libraries.

Other interfaces with the same luck are:

- `yii\di\FactoryInterface`
- `yii\rbac\CheckAccessInterface`
- maybe interface extracted from `yii\base\Application`?

But there are more. I just can't remember right now.

I think `yiisoft/interfaces` package could be created for such interfaces.

kids-return:
I am concerned that yii3.0 has been around for a long time.
I see fewer and fewer people participating in and following the community of `yii`.
There may be half a year to discuss the outcome of a problem and lead to no progress.
Ordinary developers can't get involved, this is what we need to reflect on.
Let me feel that `yii` seems to be dead.
I also made a choice between `yii` and 'laravel'.
But I really can't give up `yii`.
I hope yii is getting better and better.
About the development of yii3.0 I have some personal suggestions:
```
editing...
1.
2.
3.
4.
```

rob006:
> * allow using the package independently

`view` package already depends on `core`. You cannot use `view` without `core`, since `view` classes uses some core classes as base.

> * use other packages without the package.

This is not an argument to create separate package. There are always classes which will not be used by your app, but these will not hurt you in any way. What may hurt you, is unnecessary dependencies complexity.

The only valid reason for separate package is separate versioning, so you can introduce BC in `view` package without major release of `core`. But I'm not sure if it is worth the trouble.

kids-return:
china php framework job TOP:
```
1.ThinkPHP (Chinese development is easy to use)
2.Yii (More because yii1.0,Yii2.0 Because bower and npm are very inconvenient to install, they give up)
3.Laravel (Is becoming the second)
```

samdark:
@kids-return I understand you're worried but don't worry too much. We need time to check code before merging it.

hiqsol:
> `view` package already depends on `core`. You cannot use `view` without `core`, since `view` classes use some core classes as base.

The idea is that it can be changed.
This is already changed for: `di`, `rbac` and `mutex`. They are really usable without the core.
`log` and `cache` are close too.
I think the `view` package deserves to be freed of the core and made usable independently.
It requires work of course. But it can be done later. At the moment I think it is enough to declare intentions. When people will really need a really independent view package they will finish the work.

This is just my vision. It still can be discussed to move the `view` package into the core.

rob006:
@hiqsol I would love to see some real use case and POC for using view component without Yii core. Right now we don't know how to separate these packages and we're not sure if it makes sense at all, but we're trying to solve problems which were created by such separation. Honestly, I don't really believe that someone will use this component outside of Yii in scale that would justify additional efforts to keep this package core-independent. Especially if making `view` independent from `core` will make `view` less convenient for Yii users.

But above all IMO removing view from core makes core incomplete. I know that Yii may be used in many contexts, but it is mostly an MVC framework, and now we're removing `view` part from it...

samdark:
We're not removing, we're decoupling.

samdark:
Merged. Thanks!

hiqsol:
@rob006 This is about what is better good abstractions with poor implementation or good implementation of poor abstractions. 
It is obvious that implementation is poor at the moment and can not be fixed quickly but it is working and can be improved gradually.
I wanted the distant goal (framework made of reusable libraries) to be visible now.
It will be much harder to split out packages after releasing 3.0.

samdark:
It's not final abstraction though. Depending on implementation difficulties we may adjust goals in the process but @hiqsol is absolutely correct about intention.

rob006:
> This is about what is better good abstractions with poor implementation or good implementation of poor abstractions.

This is not the case. How can you know that this abstraction is good if it does not work as intended and you don't know how to fix it? For me it is perfect example of over-abstracting - there is no real need for separate package, there is no real problem which this will solve. It just introduces additional complicity and problems, which will make Yii less flexible (just look at this PR - now Yii does not support using views in console app, just because you want to separate `view` package) and harder to maintain.

>  It will be much harder to split out packages after releasing 3.0.

It will be much easier to do this on releasing 4.0, when there will be more time and experience gained while decoupling other packages.

hiqsol:
> It will be much easier to do this on releasing 4.0

So it is over-abstracting in 3.0 and not in 4.0 ? ;)

Ok, it was said enough, I'll stick to any yii-team decision.

rob006:
> So it is over-abstracting in 3.0 and not in 4.0 ? ;)

Just to be clear - I'm not in favor to do this in 4.0 (let say that I'm neutral in this case - it is too early say anything about 4.0 architecture). But if we don't know how to do this in 3.0 and we don't have enough time to do this right, then don't do this in 3.0. 3.0 has already so many changes that migration from 2.0 will be really hard, there is not need to rush with other BC breaking changes to make it even harder.

hiqsol:
Yes and yes.
All these considerations were already discussed between yii-team members and the decision was taken to split as much as possible to put all big BC-breaking change into 3.0 to simplify moving later releases. And clean it up eventually.
The decision can be changed though.

samdark:
Let's raise this question again closer to release.

Issue #37, opened 2018-12-28
Call to a member function allows() on array in AccessControl.php:122


terabytesoftw:
### What steps will reproduce the problem?
When executing logout in applications

### What is the expected result?
Logout

### Additional info

```
2018-12-28 18:12:05 [190.142.158.89][1][l1k039mhme6vo5jnul7hbkf3h5][error][Error] Error: Call to a member function allows() on array in /home/domain.com/public_html/vendor/yiisoft/yii-web/src/filters/AccessControl.php:122
Stack trace:
#0 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/ActionFilter.php(77): yii\web\filters\AccessControl->beforeAction(Object(yii\base\InlineAction))
#1 [internal function]: yii\base\ActionFilter->beforeFilter(Object(yii\base\ActionEvent))
#2 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/Component.php(642): call_user_func(Array, Object(yii\base\ActionEvent))
#3 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/Controller.php(276): yii\base\Component->trigger(Object(yii\base\ActionEvent))
#4 /home/domain.com/public_html/vendor/yiisoft/yii-web/src/Controller.php(367): yii\base\Controller->beforeAction(Object(yii\base\InlineAction))
#5 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/Controller.php(158): yii\web\Controller->beforeAction(Object(yii\base\InlineAction))
#6 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/Module.php(542): yii\base\Controller->runAction('logout', Array)
#7 /home/domain.com/public_html/vendor/yiisoft/yii-web/src/Application.php(94): yii\base\Module->runAction('site/logout', Array)
#8 /home/domain.com/public_html/vendor/yiisoft/yii-core/src/base/Application.php(544): yii\web\Application->handleRequest(Object(yii\web\Request))
#9 /home/domain.com/public_html/public/index.php(11): yii\base\Application->run()
#10 /home/domain.com/public_html/public/index.php(12): {closure}()
#11 {main}
```

```
Fix Instance::ensure, but it still did not work.

    public function init(): void
    {
        parent::init();
        if ($this->user !== false) {
            $this->user = Yii::ensureObject($this->user, User::class);
        }
        foreach ($this->rules as $i => $rule) {
[...]

PR #38, opened 2018-12-28
Refactored for Yii 3.0 support

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Fixed issues  | #37

### DIFF
--- a/src/filters/AccessControl.php
+++ b/src/filters/AccessControl.php
@@ -7,10 +7,10 @@
 
 namespace yii\web\filters;
 
-use yii\helpers\Yii;
 use yii\base\Action;
 use yii\base\ActionFilter;
-use yii\di\Instance;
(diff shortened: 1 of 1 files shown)


hiqsol:
Thank you! What you propose is a working solution.
But please look at how it should be done a new way.
I hope you'll catch the ideas: use DI, try not to use init and Yii helper.
That is not an ideal solution too. (And not tested)

hiqsol:
No, in general case no need for Application in model.

terabytesoftw:
it must be implemented, or it is already done.

machour:
Why would you need $this->app in models ? For translations ?

terabytesoftw:
@machour, we need $this->app and $this->getApp() in models.

Issue #11, opened 2017-12-30
Support PSR-15 HTTP Middleware


klimov-paul:
Introduce the suport for PSR-15 HTTP Middleware.

Proposed PSR draft:
https://github.com/php-fig/fig-standards/tree/master/proposed/http-handlers

Unofficial, but widely-used, PSR implementation is the `http-interop/http-server-middleware`:
https://github.com/http-interop/http-server-middleware

In order to provide the support there should be abilty to setup handlers matching the `Psr\Http\Server\MiddlewareInterface`:

```php
interface MiddlewareInterface
{
    /**
     * Process an incoming server request and return a response, optionally delegating
     * response creation to a handler.
     *
     * @param ServerRequestInterface $request
     * @param RequestHandlerInterface $handler
     *
     * @return ResponseInterface
     */
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface;
}
```

PR #40, opened 2019-01-10
Allow to prepare response outside of Request

Alex-Bond:
Hi!

This change required to receive body data for PSR-7 response passing (not echo).

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -1031,7 +1031,7 @@ class Response extends \yii\base\Response implements ResponseInterface
      * The default implementation will convert [[data]] into [[content]] and set headers accordingly.
      * @throws InvalidConfigException if the formatter for the specified format is invalid or [[format]] is not supported
      */
-    protected function prepare()
+    public function prepare()
     {
         if ($this->bodyRange !== null) {
             return;
--- a/src/Response.php
+++ b/src/Response.php
@@ -328,19 +328,29 @@ class Response extends \yii\base\Response implements ResponseInterface
 
     /**
      * Sends the response to the client.
+     * @param bool $return Send to stdout or return object. If set to `true` method will return prepared Response object.
+     * @return Response
+     * @throws HeadersAlreadySentException
+     * @throws InvalidConfigException
      */
-    public function send()
+    public function send($return = false)
     {
         if ($this->isSent) {
-            return;
+            return $this;
         }
+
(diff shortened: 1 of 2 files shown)


src/Response.php
```diff
@@ -1031,7 +1031,7 @@ public function getIsEmpty()
      * The default implementation will convert [[data]] into [[content]] and set headers accordingly.
      * @throws InvalidConfigException if the formatter for the specified format is invalid or [[format]] is not supported
      */
-    protected function prepare()
+    public function prepare()
```
samdark: The API with this change doesn't look good. How about calling it in a getter instead?

src/Response.php
```diff
@@ -328,19 +328,29 @@ public function setContent($content)
 
     /**
      * Sends the response to the client.
+     * @param bool $return Send to stdout or return object. If set to `true` method will return prepared Response object.
+     * @return Response
+     * @throws HeadersAlreadySentException
+     * @throws InvalidConfigException
      */
-    public function send()
+    public function send($return = false)
```
rob006: Now we have `send()` that does not send. :D

samdark: How about `getData()`?

Alex-Bond: It's a little confusing. Technically you send it but not in stdout but to the variable.

Alex-Bond: Like with print_r function.

Alex-Bond: Also logically it suppose to call events that you already have in send() when i requesting filled response...

samdark: Aha. Then maybe `process()` that returns data? Then `send()` will look like `echo process()`.

Alex-Bond: @samdark that can work. What about events `EVENT_BEFORE_SEND` and `EVENT_AFTER_SEND`? Technically we sending content somewhere. Or these events apply only to stdout print?

samdark: Sending events should be called for sending only. We can add two additional events for processing.  Roadrunner-like server and traditional server are two similar but different models so there may be need to do things a bit differently.

samdark:
#11

samdark:
Goind to implement proper PSR usage.

PR #79, opened 2019-06-22
Fix #68: Implement redirect

### LINKED ISSUE #68: Routing Redirect
Would you consider adding to urlManager 301/302 redirect support for the rules. This way it doesn't have to go to a controller/action of something like a deprecated path to be redirected by that action.
```
[
   'pattern' => 'some/old/route/',
   'redirect' => 'some/new/route/', // could accept parameters like how the route prop does
   'statusCode' => 301 // optional, but defaults to 301
]
```

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no tests :(
| Fixed issues  | #68

### DIFF
--- a/src/Middleware/Redirect.php
+++ b/src/Middleware/Redirect.php
@@ -60,8 +60,7 @@ final class Redirect implements MiddlewareInterface
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
         if ($this->route === null && $this->uri === null) {
-            // TODO: should we throw exception instead?
-            return $handler->handle($request);
+            throw new \InvalidArgumentException('Either toUrl() or toRoute() should be used.');
         }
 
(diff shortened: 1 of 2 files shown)


src/Middleware/Redirect.php
```diff
@@ -0,0 +1,75 @@
...
+    public function temporary(): self
+    {
+        $this->statusCode = self::TEMPORARY;
+        return $this;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if ($this->route === null && $this->uri === null) {
+            // TODO: should we throw exception instead?
```
machour: I think we should throw an exception here, unless there's a more PSR way of handling it (return a 501 ?)

Issue #90, opened 2019-07-10
Add curl command to error page


samdark:
See https://forum.yiiframework.com/t/error-handler-is-back/126667/2?u=samdark

alexkart:
@samdark , I want to write a library `curl-builder` that builds curl commands. Is it ok to use it in `yii-web` or the library must be in the `yiisoft`? See example https://github.com/yiisoft/yii-web/pull/94

samdark:
It is fine as long as you plan maintaining it properly?

samdark:
Well, not if it does what it currently does. It doesn't worth relying on external dependency for just https://github.com/alexkart/curl-builder/blob/master/src/Command.php

alexkart:
Of course, currently, it's just an example.

samdark:
Let's see how it goes. If it will have complicated logic that is usable separately — it's fine. If it will be simple enough, I'd prefer not to having another dependency.

alexkart:
Agreed, if it ends up being small I'll add the code directly to `yii-web`.

alexkart:
@samdark, so, I've finished the first version of the library, you can check it here https://github.com/alexkart/curl-builder. Is it good enough to add it to the `yii-web` as a dependency?

samdark:
Yes. I've sent a pull request: https://github.com/alexkart/curl-builder/pull/2

PR #94, opened 2019-07-23
Add curl command builder

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | 
| Fixed issues  | #90 

See comment https://github.com/yiisoft/yii-web/issues/90#issuecomment-514151141


src/ErrorHandler/templates/exception.php
```diff
@@ -411,6 +411,12 @@
         </div>
     </div>
 
+    <div class="request">
+        <div class="code">
+            <?= $this->renderCurl() ?>
```
samdark: I think it should be encoded to prevent XSS.

samdark:
Merged. Thank you!

PR #96, opened 2019-07-24
Use `Stream::__toString()` instead of `Stream::getContents`

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | 
| Fixed issues  | 

One more fix to `renderRequest()`.
Use `(string)$request->getBody()` instead of `$request->getBody()->getContents()` as consequtive calls to `getContents()` return empty string without `$request->getBody()->seek(0)`
See:
https://github.com/Nyholm/psr7/blob/master/tests/RequestTest.php#L40
https://github.com/Nyholm/psr7/blob/master/src/Stream.php#L102

### DIFF
--- a/src/ErrorHandler/HtmlRenderer.php
+++ b/src/ErrorHandler/HtmlRenderer.php
@@ -273,7 +273,7 @@ final class HtmlRenderer extends ThrowableRenderer
             }
         }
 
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

PR #97, opened 2019-07-25
Fix scheme detection

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | 
| Fixed issues  | 

When using `http` `$_SERVER['HTTPS']` can be not set or set to `off` and if `https` is used it can have any non-empty value.


samdark:
Merged. Thank you!

PR #99, opened 2019-08-08
Fix variables names by naming conventions

KartaviK:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


samdark:
It is good as is but thanks for contributing.

Issue #77, opened 2019-06-12
CSRF as middleware?


rustamwin:

samdark:
Yes.

terabytesoftw:
This middleware works well to generate the CSRF, but we need to have a Session Middleware, there will be an official one or one must be used for now use.

```
https://github.com/odan/session
https://github.com/odan/csrf
```

Any official suggestions,

Thks,

samdark:
https://github.com/yiisoft/yii-web/issues/87

samdark:
Session is done including a middleware.

terabytesoftw:
I managed to implement the CSRF token with a middleware, the bad thing about all the packages is that there must be an open session to pass the id, while I try it in this way but it would seem that I do a hacking, as we could implement it by saving it in the session, with the meta tags, or in a cokkie, will be extending the session middleware, which you think would be the best way.

```
    CsrfMiddleware::class => function (ContainerInterface $container) {
        $session = $container->get(SessionInterface::class);

        // The CSRF middleware requires a valid session id
        if (!$session->isActive()) {
            $session->open();
        }

        $csrf = new CsrfMiddleware(new Psr17Factory(), $session->getId());
        $session->set('_csrf', $csrf->getToken());

        return $csrf;
    },
```

samdark:
Session is automatically opened if you are trying to read/write from it so there's no need for explicit check.

Can't we require Session and PSR-17 factory as dependencies of CsrfMiddleware? Then there will be no need for complicated initialization.

rustamwin:
Logic should be ported from Yii2 ?

samdark:
Partially. Overall that would be kind of different because there were no middleware in 2.0.

rustamwin:
Sent PR #103

terabytesoftw:
> Session is automatically opened if you are trying to read/write from it so there's no need for explicit check.
> 
> Can't we require `Session` and PSR-17 factory as dependencies of `CsrfMiddleware`? Then there will be no need for complicated initialization then.

And in which part of the code can I access the session already open?

samdark:
1. You don't have to take care if it is open or not. Just read from it or write into it. It will open automatically.
2. You can access session anywhere via type-hinting. It is stored in DI container and will be injected.

rustamwin:
> And in which part of the code can I access the session already open?

in constructor

PR #103, opened 2019-08-11
Implement CSRF middleware

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | #77

### DIFF
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -1,0 +1,114 @@
+<?php
+
+namespace Yiisoft\Yii\Web\Middleware;
+
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Security\Random;
+use Yiisoft\Security\TokenMasker;
+use Yiisoft\Yii\Web\Cookie;
+use Yiisoft\Yii\Web\Session\SessionInterface;
+
+final class Csrf implements MiddlewareInterface
+{
+    public const CSRF_HEADER = 'X-CSRF-Token';
+
+    private $name = '_csrf';
+    private $cookieParams = ['httpOnly' => true];
+    private $enableCookie = true;
+    private $responseFactory;
+    private $session;
+
+    private function __construct(ResponseFactoryInterface $responseFactory, SessionInterface $session)
+    {
+        $this->responseFactory = $responseFactory;
+        $this->session = $session;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $token = $this->getCsrfToken($request);
+
+        if (empty($token)) {
+            $token = Random::string();
+        }
+        $request = $request->withAttribute('csrf_token', TokenMasker::mask($token));
+
+        try {
+            $response = $handler->handle($request);
+            $response = $this->addToResponse($response, $token);
+        } catch (\Throwable $e) {
+            throw new $e;
+        }
+
+        return $this->validateCsrfToken($request, $response);
+    }
+
+    private function addToResponse(ResponseInterface $response, $token): ResponseInterface
+    {
+
+        if ($this->enableCookie) {
+            $cookieParameters = \array_merge($this->session->getCookieParameters(), $this->cookieParams);
+
+            $sessionCookie = (new Cookie($this->name, $token))
+                ->path($cookieParameters['path'])
+                ->domain($cookieParameters['domain'])
+                ->httpOnly($cookieParameters['httponly'])
+                ->secure($cookieParameters['secure'])
+                ->sameSite($cookieParameters['samesite']);
+
+            return $sessionCookie->addToResponse($response);
+        } else {
+            $this->session->set($this->name, $token);
+            return $response;
+        }
+    }
+
+    private function getCsrfToken(ServerRequestInterface $request): ?string
+    {
+        if ($this->enableCookie) {
+            $cookies = $request->getCookieParams();
+            return $cookies[$this->name] ?? null;
+        } else {
+            return $this->session->get($this->name);
+        }
+    }
+
+    private function validateCsrfToken(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    {
+        $trueToken = $this->getCsrfToken($request);
+        $method = $request->getMethod();
+
+        if (\in_array($method, ['GET', 'HEAD', 'OPTIONS'], true)) {
+            return $response;
+        }
+
+        $unmaskedToken = $this->getTokenFromRequest($request);
+        if ($trueToken !== $unmaskedToken) {
+            $response = $this->responseFactory->createResponse(400);
+            $response->getBody()->write('Unable to verify your data submission.');
+            return $response;
+        }
+
+        return $response;
+    }
+
+    private function getTokenFromRequest(ServerRequestInterface $request)
+    {
+        $post = $request->getParsedBody();
+        $token = null;
+
+        if (isset($post[$this->name])) {
+            $token = $post[$this->name];
+        }
+
+        if ($token === null) {
+            $headers = $request->getHeader(self::CSRF_HEADER);
+            $token = \reset($headers);
+        }
+        return TokenMasker::unmask($token);
+    }
+}
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -32,15 +32,22 @@ final class Csrf implements MiddlewareInterface
     {
         $token = $this->getCsrfToken($request);
 
+        if (!$this->validateCsrfToken($request, $token)) {
+            $response = $this->responseFactory->createResponse(400);
+            $response->getBody()->write('Unable to verify your data submission.');
+            return $response;
+        }
+
         if (empty($token)) {
             $token = Random::string();
         }
+
         $request = $request->withAttribute('csrf_token', TokenMasker::mask($token));
 
         $response = $handler->handle($request);
         $response = $this->addTokenToResponse($response, $token);
 
-        return $this->validateCsrfToken($request, $response);
+        return $response;
     }
 
     public function setName($name)
@@ -74,23 +81,22 @@ final class Csrf implements MiddlewareInterface
         return $cookies[$this->name] ?? null;
     }
 
-    private function validateCsrfToken(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    private function validateCsrfToken(ServerRequestInterface $request, ?string $trueToken): bool
     {
-        $trueToken = $this->getCsrfToken($request);
         $method = $request->getMethod();
 
         if (\in_array($method, [Method::GET, Method::HEAD, Method::OPTIONS], true)) {
-            return $response;
+            return true;
         }
 
         $unmaskedToken = $this->getTokenFromRequest($request);
+
         if (empty($unmaskedToken) || !hash_equals($unmaskedToken, $trueToken)) {
-            $response = $this->responseFactory->createResponse(400);
-            $response->getBody()->write('Unable to verify your data submission.');
-            return $response;
+
+            return false;
         }
 
-        return $response;
+        return true;
     }
 
     private function getTokenFromRequest(ServerRequestInterface $request): ?string
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -7,7 +7,6 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
-use Yiisoft\Json\Json;
 use Yiisoft\Router\Method;
 use Yiisoft\Security\Random;
 use Yiisoft\Security\TokenMasker;
@@ -83,12 +82,7 @@ final class Csrf implements MiddlewareInterface
 
     private function getTokenFromRequest(ServerRequestInterface $request): ?string
     {
-        if ($request->getMethod() === Method::POST) {
-            $post = $request->getParsedBody();
-        } else {
-            $post = Json::decode($request->getBody(), true);
-        }
-
+        $post = $request->getParsedBody();
         $token = $post[$this->name] ?? null;
 
         if (empty($token)) {
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -62,9 +62,8 @@ final class Csrf implements MiddlewareInterface
         $token = $this->session->get($this->name);
         if (empty($token)) {
             $token = Random::string();
+            $this->session->set($this->name, $token);
         }
-
-        $this->session->set($this->name, $token);
 
         return $token;
     }
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -82,9 +82,9 @@ final class Csrf implements MiddlewareInterface
 
     private function getTokenFromRequest(ServerRequestInterface $request): ?string
     {
-        $post = $request->getParsedBody();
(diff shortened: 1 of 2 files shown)


src/Middleware/Csrf.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $trueToken = $this->getCsrfToken($request);
+        $method = $request->getMethod();
+
+        if (\in_array($method, ['GET', 'HEAD', 'OPTIONS'], true)) {
+            return $response;
+        }
+
+        $unmaskedToken = $this->getTokenFromRequest($request);
+        if ($trueToken !== $unmaskedToken) {
```
rustamwin: will compare with `StringHelper::compareString` after merging yiisoft/strings#12

rustamwin: see https://github.com/yiisoft/strings/pull/12#issuecomment-520318106

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+use Yiisoft\Security\Random;
+use Yiisoft\Security\TokenMasker;
+use Yiisoft\Yii\Web\Cookie;
+use Yiisoft\Yii\Web\Session\SessionInterface;
+
+final class Csrf implements MiddlewareInterface
+{
+    public const CSRF_HEADER = 'X-CSRF-Token';
+
+    private $name = '_csrf';
```
samdark: Should there be a way to change this name? If yes, setter is needed. If no, it could be a private constant.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+    private $cookieParams = ['httpOnly' => true];
```
samdark: Should there be a way to change it? If yes, setter is needed. If no, it could be a private constant.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+    private $enableCookie = true;
```
samdark: It could be removed because we do not support cookie-less sessions so we can assume that cookies are always available.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+
+            $sessionCookie = (new Cookie($this->name, $token))
+                ->path($cookieParameters['path'])
+                ->domain($cookieParameters['domain'])
+                ->httpOnly($cookieParameters['httponly'])
+                ->secure($cookieParameters['secure'])
+                ->sameSite($cookieParameters['samesite']);
+
+            return $sessionCookie->addToResponse($response);
+        } else {
```
samdark: Won't be executed ever.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    private function getCsrfToken(ServerRequestInterface $request): ?string
+    {
+        if ($this->enableCookie) {
+            $cookies = $request->getCookieParams();
+            return $cookies[$this->name] ?? null;
+        } else {
+            return $this->session->get($this->name);
```
samdark: Won't be executed ever.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    private function validateCsrfToken(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    {
+        $trueToken = $this->getCsrfToken($request);
+        $method = $request->getMethod();
+
+        if (\in_array($method, ['GET', 'HEAD', 'OPTIONS'], true)) {
```
samdark: There are `Method::` constants. Let's use these.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+        if (empty($token)) {
+            $token = Random::string();
+        }
+        $request = $request->withAttribute('csrf_token', TokenMasker::mask($token));
+
+        try {
+            $response = $handler->handle($request);
+            $response = $this->addTokenToResponse($response, $token);
+        } catch (\Throwable $e) {
+            throw new $e;
```
samdark: What's the goal of catch if we are not doing anything there?

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        return $this->validateCsrfToken($request, $response);
```
samdark: Isn't it too late to validate token after request handling is done?

src/Middleware/Csrf.php
mikehaertl: Why not `protected`? I'd prefer this for *all* framework properties. They should be easy to extend if you really have to.

samdark: You should not have to :)

mikehaertl: But what if you do? You never know what awkward requirements may exist in some projects. See, I've had this situation several times over all the years, that I had to copy a complete framework class file because I had to override a property or method that was defined as `private`. That's the worst of all things that can happen IMO. What exactly is the benefit of `private`? Should I open an issue to discuss this for Yii 3.0?

samdark: That is maintainability vs inheritance-patching issue. If you're making a method `protected` you can not remove, rename it or change its arguments ever since it would break backwards compatibility thus greatly limiting ability of framework maintainers to improve things in minor releases. Therefore I prefer intentional extension/configuration points instead of making everything part of the API.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,115 @@
...
+        $trueToken = $this->getCsrfToken($request);
+        $method = $request->getMethod();
+
+        if (\in_array($method, ['GET', 'HEAD', 'OPTIONS'], true)) {
+            return $response;
+        }
+
+        $unmaskedToken = $this->getTokenFromRequest($request);
+        if (empty($unmaskedToken) || !hash_equals($unmaskedToken, $trueToken)) {
+            $response = $this->responseFactory->createResponse(400);
```
rustamwin: Not sure if status code should be 400 or 403

samdark: 400 is fine.

src/Middleware/Csrf.php
rustamwin: What do you think, should be or not?

samdark: This particular parameter I think should not be changed ever.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $request = $request->withAttribute('csrf_token', TokenMasker::mask($token));
+
+        $response = $handler->handle($request);
+        $response = $this->addTokenToResponse($response, $token);
+
+        return $response;
+    }
+
+    public function setName($name)
```
samdark:
```suggestion
    public function setName(string $name): void
```

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        $this->name = $name;
+    }
+
+    public function setCookieParams($params)
```
samdark:
```suggestion
    public function setCookieParams(array $params): void
```

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        $this->cookieParams = $params;
+    }
+
+    private function addTokenToResponse(ResponseInterface $response, $token): ResponseInterface
```
samdark:
```suggestion
    private function addTokenToResponse(ResponseInterface $response, string $token): ResponseInterface
```

src/Middleware/Csrf.php
rustamwin: done

src/Middleware/Csrf.php
rustamwin: done

src/Middleware/Csrf.php
rustamwin: done

samdark:
Reviewed it in depth today. While double-submit approach with cookies works most of the time, I'd like to avoid it and stick to cookie-less synchronizer approach instead.

That is how it should work:

csrf

Adding token to HTML of all forms and adding meta tag to the page should be done transparently. The middleware will wrap all other middlewares:

```php
{
    // validate request token
    // generate tokens
    // add tokens to response HTML if content type is HTML
    return $response;
}
```

samdark:
@rustamwin do you have time to correct the pull request?

samdark:
@rob006 you have commented recently on Html helper regarding CSRF tokens and HTML modification. I'd like to hear your opinion on the approach above.

rob006:
I'm not sure if I understand, but if you're suggesting of parsing generated HTML and adding input fields with CSRF token, then it sounds quite expensive and error prone. Wouldn't be easier to create some `CsrfTokenProvider`  and pass it as argument to `Form`?

rustamwin:
> @rustamwin do you have time to correct the pull request?

I think yes. I'll do it tomorrow

samdark:
@rob006 it's not that expensive. Yii 1 used similar approach without significant performance hit.

Yii 2 switched to [replacing named placeholders](https://github.com/yiisoft/yii2/blob/master/framework/web/View.php#L82) because parsing was, indeed, error prone.

Since we don't have globally accessible repository such as `Yii::$app` we can't hide passing `CsrfTokenProvider` around. It would potentially complicate setting things up.

rob006:
@samdark I think that injecting input to forms is far more complicated than finding `<title>` or `<body>` tag. But even if we handle it, I don't feel that injecting input fields to HTML should be CsrfMiddleware responsibility (validating input and generating form are two completely different things).

samdark:
I am currently viewing it as complete self-contained universal CSRF protection middleware. The main pro is that you don't have to configure anything.

`CsrfTokenProvider` would certainly work and it will be more reliable but you have to pass provider instance around.

mikehaertl:
> CsrfTokenProvider would certainly work and it will be more reliable but you have to pass provider instance around and call getToken() explicitly in forms and in meta-tag.

FWIW If this means we'd have to manually render the CSRF hidden field in forms - I'd find this very acceptable. IMO Yii 2 often tried to do things "magically" - which may work well in 98% of the cases. But it can give you real headaches in the remaining 2% of complex scenarios.

I think it's better keep things clear and obvious even if it means that you need 1 more line to type and loose some automatism.

rob006:
@samdark it could be much simpler if we replace static helpers for generating HTML with instantiable components with reference to `View` object. Then `CsrfTokenProvider` will need to be passed only to `View` and all HTML generators will be able to use it (`$this->view->getCsrfTokenProvider()`).

samdark:
Likely yes.

rustamwin:
What about this line:
```php
$request->withAttribute('csrf_token', TokenMasker::mask($token))
```
then in controllers will able to use it and send to view

samdark:
@rustamwin great idea.

rustamwin:
@samdark everything is ok now?

samdark:
No. See https://github.com/yiisoft/yii-web/pull/103#issuecomment-521243212. The approach described doesn't use cookies at all.

rustamwin:
But did you say https://github.com/yiisoft/yii-web/pull/103#discussion_r312799238

samdark:
I did before re-checking everything in detail again.

rustamwin:
So, token should be stored only session?

samdark:
Yes.

rustamwin:
Done

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->requestName = $name;
+    }
+
+    private function getToken(): ?string
+    {
+        $token = $this->session->get($this->name);
+        if (!empty($token)) {
+            return $token;
+        } else {
+            return Random::string();
```
samdark: We need to save token to session here.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,105 @@
...
+            $response->getBody()->write('Unable to verify your data submission.');
+            return $response;
+        }
+
+        $this->session->set($this->name, $token);
+
+        $request = $request->withAttribute($this->requestName, TokenMasker::mask($token));
+        $response = $handler->handle($request);
+
+        return $response;
```
samdark: We can remove `$response` variable since we aren't doing anything with it.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $method = $request->getMethod();
+
+        if (\in_array($method, [Method::GET, Method::HEAD, Method::OPTIONS], true)) {
+            return true;
+        }
+
+        $unmaskedToken = $this->getTokenFromRequest($request);
+
+        if (empty($unmaskedToken) || !hash_equals($unmaskedToken, $trueToken)) {
```
samdark:
```php
return !empty($unmaskedToken) && hash_equals($unmaskedToken, $trueToken);
```

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+    }
+
+    private function getTokenFromRequest(ServerRequestInterface $request): ?string
+    {
+        $post = $request->getParsedBody();
+
+        $token = null;
+        if (isset($post[$this->name])) {
+            $token = $post[$this->name];
```
samdark:
```suggestion
$token = $post[$this->name] ?? null;
```

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,105 @@
...
+        if (empty($unmaskedToken) || !hash_equals($unmaskedToken, $trueToken)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    private function getTokenFromRequest(ServerRequestInterface $request): ?string
+    {
+        $post = $request->getParsedBody();
```
samdark: That is not only POST. Could be PUT or other method.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return $token;
+    }
+
+    private function validateCsrfToken(ServerRequestInterface $request, ?string $trueToken): bool
+    {
+        $method = $request->getMethod();
+
+        if (\in_array($method, [Method::GET, Method::HEAD, Method::OPTIONS], true)) {
```
rustamwin: Hmm ... Shouldn't we regenerate the token every request like Yii2? @samdark

src/Middleware/Csrf.php
rustamwin: need review my approach

src/Middleware/Csrf.php
samdark: No. That does not give you anything security-wise but may affect usability for multiple tabs browsing.

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,101 @@
...
+
+        return !empty($unmaskedToken) && hash_equals($unmaskedToken, $trueToken);
+    }
+
+    private function getTokenFromRequest(ServerRequestInterface $request): ?string
+    {
+        if ($request->getMethod() === Method::POST) {
+            $post = $request->getParsedBody();
+        } else {
+            $post = Json::decode($request->getBody(), true);
```
samdark: I don't think it is a good idea to assume non-POST should contain JSON-body. Using `getParsedBody()` is OK in all cases. It was variable name I was complained about.

rustamwin: :)

src/Middleware/Csrf.php
```diff
@@ -0,0 +1,95 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Router\Method;
+use Yiisoft\Security\Random;
+use Yiisoft\Security\TokenMasker;
+use Yiisoft\Yii\Web\Session\SessionInterface;
+
+final class Csrf implements MiddlewareInterface
+{
+    public const NAME = '_csrf';
```
samdark: Are these constants useful outside of the class?

rustamwin: I think yes, but besides this

samdark:
Merged. Thank you very much!

PR #111, opened 2019-09-03
update phpdocs

iamsaint:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


samdark:
Thank you for participating but this pull request won't be merged because in this case phpdoc does not add any extra information that is not there yet in the method signatures.

PR #120, opened 2019-10-04
Adjusted public API of Flash

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Flash.php
+++ b/src/Flash.php
@@ -20,15 +20,13 @@ final class Flash implements FlashInterface
         $this->session = $session;
     }
 
-    public function get(string $key, $defaultValue = null)
+    public function get(string $key)
     {
         $flashes = $this->fetch();
 
         if (!isset($flashes[$key], $flashes[self::COUNTERS][$key])) {
-            return $defaultValue;
+            return null;
         }
-
-        $value = $flashes[$key];
 
         if ($flashes[self::COUNTERS][$key] < 0) {
             // mark for deletion in the next request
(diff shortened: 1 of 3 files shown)


samdark:
@pchapl what do you think?

src/Flash.php
```diff
@@ -89,21 +85,35 @@ public function add(string $key, $value = true, bool $removeAfterAccess = true):
         $this->save($flashes);
     }
 
-    public function remove(string $key)
+    public function remove(string $key, $defaultValue = null)
```
rob006: `remove()` with `$defaultValue` parameter = WTF! :P

samdark: That is to match `get()`. Remove returns flash message.

rob006: This is optimization for reading value. This means that either this method is badly named (the main purpose of it is to get value - removing value from collection is only additional side effect) or default value logic does not belong here and will be only confusing.

IMO it is both of them. I never liked `ArrayHelper::remove()` for exact the same reason - it is confusing to use `remove()` method for reading value from config array. I'm always trying to extend this helper with more adequate name, which will also imply getting value (like `pop()`).
And since `??` operator supporting default value is mostly pointless - `??` is more readable and more efficient.

pchapl:
I agree with @rob006 :
1. default value is unnecessary in all methods;
2. `remove` and `removeAll` should be void.

However, in my opinion, removing the `$delete` parameter is a good idea.

samdark:
Agree. Made adjustments.

pchapl:
Looks good 👍

Issue #119, opened 2019-10-02
Network resolver instead of operate with REMOTE ADDR and more


kamarton:
I see the middleware already contains IP filtering so I would raise the following issues and suggestions. 

The following examples are based on the use of yii2.

### What steps will reproduce the problem?

Yii2's basic idea of ​​secure networks, headers, and IP definition is too simple.
- On the one hand, there is an unnecessary function in the Request class `trustedHosts` and `secureHeaders` when not in use.
- On the other hand, it is difficult or impossible to handle some of the networks schema.

The yii2 solution unnecessarily write conditions in the code that are difficult to redefine.

Here are some examples of network resolution:
- check IP by ISP or reverse DNS
- check route by network endpoints sequence (example from X-Forwarded-For header)
- check and skip optional route's IP, if needed
- check and skip route loops
- get IP by tagged endpoint (The multiple routes require the IP addresses of the network endpoints, which may check and get the first, the last, all and exists.)
- resolve user IP from accepted sequenced route

The former need not be integrated into the framework, but should be defined as an interface with a separate layer.

### Suggests

```php
interface NetworkResolverInterface {
  public function getRemoteIp(): string;
  public function getUserIp(): string;
  public function getIsSecureConnection(): bool;
}
```

```php
class Request ... {
   private $networkResolver;

   __construct(..) {
     $this->networkResolver = SimpleNetworkResolver(); // use REMOTE_ADDR, SERVER_PORT etc.
   }

   public function withNetworkResolver(NetworkResolver $resolver) {..}
   public function getNetworkResolver(): NetworkResolver {...}
}
```

example `TrustedHostsResolver` class is the equivalent of yii2 trusted hosts.

| Q | A
| ---------------- | ---
| Version | 1.0.0-under development
| PHP version | -
| Operating system | -

samdark:
1. I don't think `getIsSecureConnection()` belongs here. It is part of PSR-7 i.e. you can do `$request->getUri()->getScheme() === 'https'` and it should be reliable.
2. `NetworkResolverInterface` becomes `IpInformationInterface` or `IpResolverInterface` then.

kamarton:
> I don't think getIsSecureConnection() belongs here. It is part of PSR-7 i.e. you can do $request->getUri()->getScheme() === 'https' and it should be reliable.

PSR-7 OK.

```php
interface NetworkResolverInterface {
  public function getRemoteIp(): string;
  public function getUserIp(): string;
  public function getScheme(): string;
[...]

PR #125, opened 2019-10-09
NetworkResolver #119

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #119

- ~~[ ] network resolver interface~~
- [x] resolvers
  - [x] BasicNetworkResolver
    - [x] documentation
    - [x] tests
  - ~~[ ] TrustedHostsNetworkResolver~~


src/Middleware/IpFilter.php
```diff
@@ -27,7 +47,8 @@ public function __construct(string $allowedIp, ResponseFactoryInterface $respons
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        if ($request->getServerParams()['REMOTE_ADDR'] !== $this->allowedIp) {
+        $networkResolver = $this->networkResolver->withServerRequest($request);
```
samdark: `$this->networkResolver` seems to be optional so a check for `null` should be added with falling back to `$request->getServerParams()['REMOTE_ADDR']`. Alternatively it could be made mandatory.

src/NetworkResolver/NetworkResolverInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+
+namespace Yiisoft\Yii\Web\NetworkResolver;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+interface NetworkResolverInterface
+{
+    public function getRemoteIp(): string;
```
samdark: It cannot work without server request so either `ServerRequestInterface` should be injected from constructor or passed as a method argument. I prefer:

```php
 public function getRemoteIp(ServerRequestInterface $serverRequest): string;
```

src/NetworkResolver/NetworkResolverInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+    public function getUserIp(): string;
```
samdark: Same as above, I think we need to add ServerRequestInterface  here.

kamarton:
> MiddlewareDispatcher: this one will adjust Request to contain trusted info / additional info

Could `MiddlewareDispatcher` pass the network resolver to middleware objects? (`IpFilter::networkResolver`) Either need it to use it at all or will ServerRequest already contain the modified information you want in `Middleware::process()`?

src/NetworkResolver/NetworkResolverInterface.php
samdark: > else use $serverRequest from the object ($this)

Where would you get it from? It's not in the container.

samdark:
> Could MiddlewareDispatcher pass the network resolver to middleware objects?

Yes.

> (IpFilter::networkResolver) Either need it to use it at all or will ServerRequest already contain the modified information you want in Middleware::process()?

I think it should contain already modified info that could be trusted.

src/NetworkResolver/NetworkResolverInterface.php
kamarton: > Where would you get it from? It's not in the container.

```php
class AnyNetworkResolver implements... {
  private $serverRequest;
  public function withServerRequest($serverRequest) {...}
```

The object can be used multiple times during runtime. Under normal use, it must be instantiated the first time the `$serverRequest` object appears. However, the instance should be accessible anywhere where the `$serverRequest` object is accessible.

User IP usage examples:
- insert into database
- condition in controller
- display in view

Based on the tests, I can see the current architecture of the system and not see where it can be injected.

```php
return new MiddlewareDispatcher([
    $errorCatcher,
    $networkResolver, // <---- It should not be placed in a parameter here, but rather in a `$container`.
    $session,
    $router, // controller would get Request that was modified by networkResolver.
                 // No need to pass networkResolver as dependency of the widget
                 //   ^----------------- It's OK
], $container);
//     ^
//     \-------------------- Maybe we should store it here?
```

samdark: Yes, NetworkResolver could be put into container but it won't get request instance anyway. So why having `withRequest()` then? I think this method should be removed.

kamarton: > So why having withRequest() then? I think this method should be removed.

The `$request` object passed to all methods is a bit like a static helper class. I think you need to find the right point in the code where
- `$request` object already exists
- At the configuration level has already been added to the container (network resolver)

Unless a middle ware setting something in the network resolver, its output does not change during the web request. If the object internally stores the request object, it can cache the output results. However, if you get the request object in a parameter, I cannot assume that the request object will not change.

kamarton: Final result of planning:
- `Middleware\NetworkResolver`: If needed, it modifies the request
- `NetworkResolver\NetworkResolverInterface`: resolving the network. Depending on the use, may be placed in a container or helper (if needed, eg user ip).

samdark: OK.

samdark:
Seems to fail tests.

kamarton:
> Seems to fail tests.

Yes. There is a special case that is not supported by yii2: the user is connected via HTTP, but the last proxy is already using HTTPS (eg more efficient HTTP2/QUIC)

```
client <-----http----> CDN <-----http2/quic----> server
```

A test has been added, this is what is causing the error. I want to read if this implementation makes sense.

samdark:
I think such configuration is possible.

samdark:
👍 What's left?

kamarton:
>  What's left?

`TrustedHostNetworkResolver`, test separation, more tests. and minimal phpdoc for complex method's parameters.

kamarton:
@samdark `composer.lock` is ignored in git, why? The `TrustedHostsNetworkResolver` will require the `Ip` validator from the` yiisoft/validator` package.

https://github.com/yiisoft/yii-web/blob/3398f9b5ccdeea2365d51f62498f4e1ce03910a8/.gitignore#L19-L21

---
I have modified the interface because this must be able to modify the many of parameters in the request object eg in `TrustedHostsNetworkResolver`. Not least, it makes the interface simpler and more flexible.

samdark:
In order not to stuck with outdated dependencies. `composer.json` is read/evaluated every time.

src/Middleware/NetworkResolver.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Yiisoft\Yii\Web\Middleware;
+
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Yii\Web\NetworkResolver\NetworkResolverInterface;
+use Yiisoft\Yii\Web\NetworkResolver\NotTrustedHeadersInterface;
```
samdark: There's no such class in the pull request.

src/NetworkResolver/NetworkResolverInterface.php
```diff
@@ -0,0 +1,30 @@
...
+use Nyholm\Psr7\Uri;
+use Psr\Http\Message\ServerRequestInterface;
+
+interface NetworkResolverInterface
+{
+
+    /**
+     * @return static
+     */
+    public function withServerRequest(ServerRequestInterface $serverRequest);
```
samdark: I still think that it's no correct. Server request is not optional. It is required for all interface methods. So, since resolver is meant to be called with different request instances, this method and `getServerRequest()` should be removed and `$request` should be added to `getRemoveIp()`, `getUserIp()`, `isSecureConnection()`.

samdark: Current `getServerRequest()` code could be moved to middleware.

kamarton: > I still think that it's not correct. Server request is not optional.

With the PSR structure of yii3 I don't see a better solution.

> and $request should be added to getRemoveIp(), getUserIp(), isSecureConnection().

This solution is not developers friendly. And besides that, not really an OOP approach, but rather static functions.

```php
//yii2
$request->getUserIp()
//symfony: (https://symfony.com/doc/current/create_framework/http_foundation.html)
$request->getClientIp()
// symfony: (https://stackoverflow.com/questions/9029757/how-do-i-get-the-user-ip-address-in-symfony2-controller)
$this->container->get('request')->getClientIp();
public function indexAction(Request $request)
{
    $ip = $request->getClientIp();
}

//and yii3:
$container->get('networkResolver')->getUserIp($request);
```

> Current getServerRequest() code could be moved to middleware.

Modifying IPs and request is a one merged process. Separation would mean duplicate code. It might be possible to outsource the common code section, but that would only complicate the code. Can be solved with "BasicNetworkResolver", but not with the more complex IP process.

kamarton:
@samdark The `Ip` rule can't find classes. Need more entries in `composer.json`?

```
1) Yiisoft\Yii\Web\Tests\NetworkResolver\TrustedHostsNetworkResolverTest::testIps with data set #0 ('127.0.0.1', array(array('8.8.8.8, 9.9.9.9')), array('127.0.0.1'), null, '9.9.9.9')
Error: Class 'yii\helpers\IpHelper' not found

/home/smarton/work/source/web/yii2/kamarton-yii3-web/vendor/yiisoft/validator/src/Rule/Ip.php:464
/home/smarton/work/source/web/yii2/kamarton-yii3-web/vendor/yiisoft/validator/src/Rule/Ip.php:304
/home/smarton/work/source/web/yii2/kamarton-yii3-web/vendor/yiisoft/validator/src/Rule/Ip.php:260
/home/smarton/work/source/web/yii2/kamarton-yii3-web/src/NetworkResolver/TrustedHostsNetworkResolver.php:228
/home/smarton/work/source/web/yii2/kamarton-yii3-web/src/NetworkResolver/TrustedHostsNetworkResolver.php:167
/home/smarton/work/source/web/yii2/kamarton-yii3-web/src/NetworkResolver/TrustedHostsNetworkResolver.php:125
/home/smarton/work/source/web/yii2/kamarton-yii3-web/tests/NetworkResolver/TrustedHostsNetworkResolverTest.php:36
```

src/Middleware/NetworkResolver.php
kamarton: removed

src/NetworkResolver/NetworkResolverInterface.php
kamarton: > Current getServerRequest() code could be moved to middleware.

I can imagine splitting it into 2 interfaces and `NetworkResolverInterface` would implement it. And middleware is generalized, eg. `RequestModifier`. There are processes where eg. deleting or modifying headers.

kamarton: Plan B:
The network resolver middleware will have multiple instances and store the IP addresses as an attribute.
```php
// in network resolver middleware
return $request->withAttribute('clientIp', $clientIp)->withAttribute('remoteIp', $remoteIp);
```

@samdark reviews?

samdark: > //yii3 with more developer friendly
> $container->get('networkResolver')->getUserIp()

Where from would it get server request? It's not in the container.

> UPD

Is there any situation where we need both user IP and server IP at the same time? If not, I'd prefer `NetworkResolver` to modify request directly by changing values without adding any extra attributes.

samdark:
IpRule isn't ready. We can extract something like yiisoft/network package containing tools for various network calculations such as IP matching, masks, ranges etc.

src/NetworkResolver/NetworkResolverInterface.php
kamarton: > Is there any situation where we need both user IP and server IP at the same time?

I think there is a case where both are needed. Of course, this is a special use. I think that is perfectly acceptable. And if you need both in one project, you can do it with your own middleware development.

Only very uncomfortable with use.

```php
$ip = $request->getServerParams()['REMOTE_ADDR'];
```

kamarton: > Is there any situation where we need both user IP and server IP at the same time? If not, I'd prefer NetworkResolver to modify request directly by changing values without adding any extra attributes.

I start implementing it based on that and then it becomes middleware only.

kamarton: `BasicNetworkResolver` refactored to middleware is complete.

kamarton:
@samdark Should I put `TrustedHostsNetworkResolver` in a separate branch in the absence of an IP validator and delete it from here?

samdark:
> Should I put TrustedHostsNetworkResolver in a separate branch in the absence of an IP validator and delete it from here?

Yes, let's do that after this one is merged.

kamarton:
> Yes, let's do that after this one is merged.

removed

kamarton:
@samdark the code is reviewable.

src/Middleware/BasicNetworkResolver.php
```diff
@@ -0,0 +1,141 @@
...
+        foreach ($this->protocolHeaders as $header => $data) {
+            if (!$request->hasHeader($header)) {
+                continue;
+            }
+            $headerValues = $request->getHeader($header);
+            if (is_callable($data)) {
+                $newScheme = call_user_func($data, $headerValues, $header, $request);
+                if ($newScheme === null) {
+                    continue;
+                } elseif (!is_string($newScheme)) {
```
machour: the `elseif` isn't needed here, can be a simple if

src/Middleware/BasicNetworkResolver.php
```diff
@@ -0,0 +1,141 @@
...
+                    throw new \RuntimeException('The scheme is neither string nor null!');
+                } elseif (strlen($newScheme) === 0) {
```
machour: same here

src/Middleware/BasicNetworkResolver.php
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param callable|array|null $protocolAndAcceptedValues
+     * @return static
+     * @see DEFAULT_PROTOCOL_AND_ACCEPTABLE_VALUES
+     */
+    public function withProtocolHeader(string $header, $protocolAndAcceptedValues = null)
+    {
+        $new = clone $this;
+        $header = strtolower($header);
+        if ($protocolAndAcceptedValues === null) {
```
machour: I'd like to get rid of the if/elseif construction to enhance readability.
Would it be possible to return `$new` right away or throw ?

src/Middleware/BasicNetworkResolver.php
kamarton: modified

src/Middleware/BasicNetworkResolver.php
kamarton: modified

src/Middleware/BasicNetworkResolver.php
kamarton: Yes, modified.

src/NetworkResolver/NetworkResolverInterface.php
kamarton: Now I see that the `$server` parameters cannot be overwrite without a new instance (`withServerParams ()` is not defined). Regardless, the middleware implementation is correct, but it is unavoidable to use `$request->withAttributes(...)`. At the framework level, this also means that the attribute name of user's IP should be unified.

samdark:
Merged. Thank you!

Do you need a repository for IP / network helper?

kamarton:
> Do you need a repository for IP / network helper?

I think for the time being I can work it out to make the IP validator workable in local, but it will definitely be necessary for a PR.

samdark:
https://github.com/yiisoft/netutils. I'll add files there soon.

kamarton:
> https://github.com/yiisoft/netutils. I'll add files there soon.

The `yiisoft/network-utility` would not be nicer?

samdark:
https://github.com/yiisoft/network-utilities

samdark:
Package is ready.

PR #127, opened 2019-10-23
Fix #121 removed unnecessary ucwords header name

### LINKED ISSUE #121: ucwords header name
### What steps will reproduce the problem?

I think it's unnecessary to convert the first letters of the words to uppercase.

https://github.com/yiisoft/yii-web/blob/11a354d8542de7f5b4d0378b321f1226d5223f93/src/ServerRequestFactory.php#L143

This solution also causes quite a few unnecessary conversions instead of simply `strtolower()` conversions.

HTTP2 use only lower case: https://tools.ietf.org/html/rfc7540#section-8.1.2

### What is the expected result?

`HTTP_X_FORWARDED_FOR` -> `x-forwarded-for`

### What do you get instead?

`HTTP_X_FORWARDED_FOR` -> `X-Forwarded-For`

| Q                | A
| ---------------- | ---
| Version          | 1.0.0-under development
| PHP version      | -
| Operating system | -

pistej:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #121

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -140,7 +140,7 @@ final class ServerRequestFactory
             $headers = [];
             foreach ($_SERVER as $name => $value) {
                 if (strncmp($name, 'HTTP_', 5) === 0) {
-                    $name = str_replace(' ', '-', ucwords(strtolower(str_replace('_', ' ', substr($name, 5)))));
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

Issue #128, opened 2019-10-24
Invalid IPv6 host parsing


kamarton:
### What steps will reproduce the problem?

If the hostname is not specified, it uses square brackets for IPv6.

https://github.com/yiisoft/yii-web/blob/e5a541931c81721df5c360d62b4a2d2d263b3d76/src/ServerRequestFactory.php#L107-L112

![Desktop screenshot](https://user-images.githubusercontent.com/9432032/67449342-27deb480-f61a-11e9-8a6a-12ba88aeddc0.png)

### What is the expected result?

`[::1]` -> `::1`
`[::1]:443` -> `::1`

### What do you get instead?

`[::1]` -> `[::1]`
`[::1]:443` -> `[::1]`

| Q                | A
| ---------------- | ---
| Version          | 1.0.0-under-development
| PHP version      | -
| Operating system | -

kamarton:
This is not a bug.

PR #129, opened 2019-10-24
invalid IPv6 host parsing fix and tests #128

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #128


kamarton:
I move the tests to another PR, the base problem is junk.

Issue #105, opened 2019-08-19
Add tests


samdark:
Currently yii-web package is not properly tested. Need to add tests.

PR #137, opened 2019-10-30
WebActionsCaller, ActionCaller and TagRequest tests

zhukovra:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #105

### DIFF
--- a/tests/Middleware/ActionCallerTest.php
+++ b/tests/Middleware/ActionCallerTest.php
@@ -12,22 +12,34 @@ use PHPUnit\Framework\TestCase;
 
 class ActionCallerTest extends TestCase
 {
+    /** @var ServerRequestInterface  */
+    private $request;
+
+    /** @var RequestHandlerInterface  */
+    private $handler;
+
+    protected function setUp()
+    {
+        $this->request = $this->createMock(ServerRequestInterface::class);
+        $this->handler = $this->createMock(RequestHandlerInterface::class);
+    }
+
     public function testProcess()
     {
         $container = $this->createMock(ContainerInterface::class);
         $container->method('get')->willReturn($this);
 
-        $request = $this->createMock(ServerRequestInterface::class);
-        $handler = $this->createMock(RequestHandlerInterface::class);
+        $caller = new ActionCaller('this', 'process', $container);
 
-        $caller = new ActionCaller('this', 'exampleMethod', $container);
-
-        $response = $caller->process($request, $handler);
+        $response = $caller->process($this->request, $this->handler);
         self::assertEquals(204, $response->getStatusCode());
     }
 
-    public function exampleMethod(): ResponseInterface
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
+        $this->assertSame($this->request, $request);
+        $this->assertSame($this->handler, $handler);
+
         return new Response(204);
     }
 }
--- a/tests/Middleware/ActionCallerTest.php
+++ b/tests/Middleware/ActionCallerTest.php
@@ -3,10 +3,10 @@
 namespace Yiisoft\Yii\Web\Tests\Middleware;
 
 use Nyholm\Psr7\Response;
-use Psr\Container\ContainerInterface;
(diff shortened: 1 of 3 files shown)


tests/Middleware/WebActionsCallerTest.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Yiisoft\Yii\Web\Middleware;
+
+use Nyholm\Psr7\Response;
+use PHPUnit\Framework\TestCase;
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class WebActionsCallerTest extends TestCase
```
samdark: Same as in case with `ActionCallerTest` it worth checking that:

1. Request is passed to `example()`
2. Dependency injection by method signature works in `example()`

tests/Middleware/ActionCallerTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $request = $this->createMock(ServerRequestInterface::class);
+        $handler = $this->createMock(RequestHandlerInterface::class);
+
+        $caller = new ActionCaller('this', 'exampleMethod', $container);
+
+        $response = $caller->process($request, $handler);
+        self::assertEquals(204, $response->getStatusCode());
+    }
+
+    public function exampleMethod(): ResponseInterface
```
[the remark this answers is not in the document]
zhukovra: Without passing to exampleMethod `assertEquals(204, $response->getStatusCode());` fails, because not default 204 response code sets in exampleMethod(). DI returns ActionCallerTest itself and ActionCaller calls 'exampleMethod' in ActionCallerTest instance.

Should I rewrite all on mocks without using "helper" methods of test classes?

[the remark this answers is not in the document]
samdark: I've meant checking that `$request` object is the same instance as was mocked and that `$test` is an instance of `CustomDependency` obtained from container:

```php
public function exampleMethod(ServerRequestInterface $request, CustomDependency $test): ResponseInterface
{
   // ...
}
```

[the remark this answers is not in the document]
samdark: > Should I rewrite all on mocks without using "helper" methods of test classes?

The less mocks the better.

tests/Middleware/ActionCallerTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $container = $this->createMock(ContainerInterface::class);
+        $container->method('get')->willReturn($this);
+
+        $caller = new ActionCaller('this', 'process', $container);
+
+        $response = $caller->process($this->request, $this->handler);
+        self::assertEquals(204, $response->getStatusCode());
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
samdark: Now it verifies that request and handler are correctly passed 👍

Let's expand it to type-hinting for a custom service to verify that this service will be searched in DI container.

tests/Middleware/WebActionsCallerTest.php
```diff
@@ -0,0 +1,74 @@
...
+            ->willReturn('notExistant');
+
+        $this->handler
+            ->expects($this->once())
+            ->method('handle');
+
+        (new WebActionsCaller('this', $this->container))->process($this->request, $this->handler);
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
samdark: Same as in comment above.

tests/Middleware/ActionCallerTest.php
zhukovra: Is it acceptable changes?
image

samdark: Not mocking container? Yes. Absolutely. That's actually way cleaner than using a mock.

samdark:
Merged. Thank you!

Issue #149, opened 2019-11-06
ErrorCatcher ignore q factor from accept header items


kamarton:
### What steps will reproduce the problem?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept#Directives

https://github.com/yiisoft/yii-web/blob/63284046693a2206ba232f7f10c3f93929dcbc70/src/Middleware/ErrorCatcher.php#L66

### What is the expected result?

Factor q is also considered when selecting a renderer.

### What do you get instead?

Disregards factor q.

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

PR #151, opened 2019-11-11
ErrorCatcher with RFC7231 accept header process #149

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #149

### DIFF
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -32,7 +32,7 @@ final class HeaderHelper
      * @see getValueAndParameters
      * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
      */
-    public static function getByQFactorSortedList($values, bool $outputValuesOnly = false): array
+    public static function getByQFactorSortedList($values): array
     {
         if (is_string($values)) {
             $values = preg_split('/\s*,\s*/', trim($values), -1, PREG_SPLIT_NO_EMPTY);
@@ -48,7 +48,7 @@ final class HeaderHelper
             $parse = self::getValueAndParameters($value);
             $q = $parse['q'] ?? 1.0;
             if (is_string($q) && preg_match('/^(?:0(?:\.\d{1,3})?|1(?:\.0{1,3})?)$/', $q) === 0) {
-                throw new \InvalidArgumentException('Invalid q factor!');
+                throw new \InvalidArgumentException('Invalid q factor');
             }
             $parse['q'] = floatval($q);
             $output[] = $parse;
@@ -61,12 +61,6 @@ final class HeaderHelper
             }
             return $a > $b ? -1 : 1;
         });
-        if (!$outputValuesOnly) {
-            return $output;
-        }
-        foreach ($output as &$value) {
-            $value = array_shift($value);
-        }
         return $output;
     }
 
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -14,7 +14,7 @@ final class HeaderHelper
     public static function getValueAndParameters(string $headerValue): array
     {
         $headerValue = trim($headerValue);
-        if (strlen($headerValue) === 0) {
+        if ($headerValue === '') {
             return [];
         }
         $parts = preg_split('/\s*;\s*/', $headerValue, -1, PREG_SPLIT_NO_EMPTY);
@@ -38,7 +38,7 @@ final class HeaderHelper
             $values = preg_split('/\s*,\s*/', trim($values), -1, PREG_SPLIT_NO_EMPTY);
         }
         if (!is_array($values)) {
-            throw new \InvalidArgumentException('Values ​​are neither array nor string!');
+            throw new \InvalidArgumentException('Values ​​are neither array nor string');
         }
         if (count($values) === 0) {
             return [];
@@ -53,7 +53,7 @@ final class HeaderHelper
             $parse['q'] = floatval($q);
             $output[] = $parse;
         }
-        usort($output, function ($a, $b) {
+        usort($output, static function ($a, $b) {
             $a = $a['q'];
             $b = $b['q'];
             if ($a === $b) {
@@ -84,7 +84,7 @@ final class HeaderHelper
             $values = $values->getHeader('accept');
         }
         $output = self::getByQFactorSortedList($values);
-        usort($output, function ($a, $b) {
+        usort($output, static function ($a, $b) {
             if ($a['q'] !== $b['q']) {
                 // The higher q value wins
                 return $a['q'] > $b['q'] ? -1 : 1;
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -32,7 +32,7 @@ final class HeaderHelper
      * @see getValueAndParameters
      * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
      */
-    public static function getByQFactorSortedList($values): array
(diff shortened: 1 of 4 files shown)


src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+{
+    /**
+     * Explode header value to value and parameters (eg. text/html;q=2;version=6)
+     * @param string $headerValue
+     * @return array first element is the value, and key-value are the parameters
+     */
+    public static function getValueAndParameters(string $headerValue): array
+    {
+        $headerValue = trim($headerValue);
+        if (strlen($headerValue) === 0) {
```
samdark:
```suggestion
        if ($headerValue === '') {
```

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+     * @see getValueAndParameters
+     * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+     */
+    public static function getByQFactorSortedList($values, bool $outputValuesOnly = false): array
+    {
+        if (is_string($values)) {
+            $values = preg_split('/\s*,\s*/', trim($values), -1, PREG_SPLIT_NO_EMPTY);
+        }
+        if (!is_array($values)) {
+            throw new \InvalidArgumentException('Values ​​are neither array nor string!');
```
samdark:
```suggestion
            throw new \InvalidArgumentException('Values ​​are neither array nor string');
```

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+        return $output;
+    }
+
+    /**
+     * Getting header value as q factor sorted list
+     * @param string|string[] $values Header value as a comma-separated string or already exploded string array.
+     * @see getValueAndParameters
+     * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+     */
+    public static function getByQFactorSortedList($values, bool $outputValuesOnly = false): array
```
samdark:
```suggestion
    public static function getSortedValueAndParameters($values, bool $outputValuesOnly = false): array
```

samdark: Name is longer but, I think, is a bit more readable.

samdark: It's not clear to me what is expected as input value for `$values`.

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+        foreach ($values as $value) {
+            $parse = self::getValueAndParameters($value);
+            $q = $parse['q'] ?? 1.0;
+            if (is_string($q) && preg_match('/^(?:0(?:\.\d{1,3})?|1(?:\.0{1,3})?)$/', $q) === 0) {
+                throw new \InvalidArgumentException('Invalid q factor!');
+            }
+            $parse['q'] = floatval($q);
+            $output[] = $parse;
+        }
+        usort($output, function ($a, $b) {
```
samdark:
```suggestion
        usort($output, static function ($a, $b) {
```

src/Helper/HeaderHelper.php
samdark: Also I don't like `$outputValuesOnly`. It looks like method is doing too much and could be actually split into two more specific and simpler methods (or more).

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+            return $output;
+        }
+        foreach ($output as &$value) {
+            $value = array_shift($value);
+        }
+        return $output;
+    }
+
+    /**
+     * @param $values string|string[]|RequestInterface $values Header value as a comma-separated string
```
samdark: It is better to have at least a separate method that works with RequestInterface. Having too many types in a single parameter adds to overall method complexity including its usage.

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,128 @@
...
+     *                  added to the type, which are always appended by a semicolon and sorted by string.
+     * @link https://tools.ietf.org/html/rfc7231#section-5.3.2
+     */
+    public static function getSortedAcceptTypes($values): array
+    {
+        if ($values instanceof RequestInterface) {
+            $values = $values->getHeader('accept');
+        }
+        $output = self::getByQFactorSortedList($values);
+        usort($output, function ($a, $b) {
```
samdark:
```suggestion
        usort($output, static function ($a, $b) {
```

src/Middleware/ErrorCatcher.php
```diff
@@ -99,7 +100,7 @@ private function getRenderer(string $contentType): ?ThrowableRendererInterface
 
     private function getContentType(ServerRequestInterface $request): string
     {
-        $acceptHeaders = preg_split('~\s*,\s*~', $request->getHeaderLine('Accept'), PREG_SPLIT_NO_EMPTY);
+        $acceptHeaders = HeaderHelper::getSortedAcceptTypes($request);
```
samdark: It is the only usage of the helper so far. What do you think about making some of the methods of the helper private?

src/Helper/HeaderHelper.php
kamarton: > Also I don't like `$outputValuesOnly`. It looks like method is doing too much and could be actually split into two more specific and simpler methods (or more).

Really, this parameter is not required. Later I saw that I can't use it in to accept the header, because the process itself is so complicated. In my opinion, post-processing is generally required, so this parameter is not relevant.

kamarton: > Name is longer but, I think, is a bit more readable.

I check to see if there is any header content that is sorted differently, because in that case the method name might be problematic.

src/Middleware/ErrorCatcher.php
kamarton: It may be problematic in terms of testability.

src/Helper/HeaderHelper.php
```diff
@@ -0,0 +1,122 @@
...
+        return $output;
+    }
+
+    /**
+     * Getting header value as q factor sorted list
+     * @param string|string[] $values Header value as a comma-separated string or already exploded string array.
+     * @see getValueAndParameters
+     * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+     */
+    public static function getByQFactorSortedList($values): array
```
samdark: getValuesListSortedByQFactor?

src/Middleware/ErrorCatcher.php
samdark: Why? Tests should reflect usage overall.

kamarton: > Why? Tests should reflect usage overall.

This is true. The problem is that e.g. testing the `A` and `B` functions that are built into the `C` function is usually tested with fewer tests than the `A`, `B`, and `C` functions are tested separately. Obviously this is wrong, but usually it will be the result.

src/Helper/HeaderHelper.php
kamarton: How about using only `RequestInterface` as input?

src/Helper/HeaderHelper.php
kamarton: `getSortedValueAndParameters`

No other sorting method found. What I have found is all sorted by q factor, so it's unnecessary to name q factor in the method name.

samdark:
Merged. Thank you!

PR #153, opened 2019-11-15
HeaderHelper:getValueAndParameters() quote fix #149

### LINKED ISSUE #149: ErrorCatcher ignore q factor from accept header items

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -

https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html

HTTP 1 1  Protocol Parameters

HTTP 1 1  Protocol Parameters

https://tools.ietf.org/html/rfc7239#section-6

> It is important to note that an IPv6 address and any nodename with
   node-port specified MUST be quoted, since ":" is not an allowed
   character in "token".

https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

RFC1341 MIME    7 The Multipart content type

### DIFF
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -21,6 +21,9 @@ final class HeaderHelper
         $output = [array_shift($parts)];
         foreach ($parts as $part) {
             [$key, $headerValue] = explode('=', $part, 2);
+            if (preg_match('/^"(?<value>.*)"$/', $headerValue, $matches) === 1) {
+                $headerValue = $matches['value'];
+            }
             $output[$key] = $headerValue;
         }
         return $output;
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

PR #157, opened 2019-11-20
Trusted hosts network resolver

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #119

- [x] more tests
- [x] more method comments


kamarton:
@samdark please pre check the code. Based on a few tests, the code is already functional.

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,445 @@
...
+        return $validator->ranges($ranges)->validate($host)->isValid();
+    }
+
+    /**
+     * Reverse obfuscating host data
+     *
+     * The base operation does not perform any transformation on the data.
+     * This method can be extendable by overwriting eg.
+     */
+    protected function reverseObfuscate(array $ipData, array $ipDataList): array
```
samdark: `$ipDataList` isn't used. Is it here intentionally?

samdark: Method is protected but it's not overwritten anywhere. Is it meant to be extended?

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,446 @@
...
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\NetworkUtilities\IpHelper;
+use Yiisoft\Validator\Rule\Ip;
+use Yiisoft\Yii\Web\Helper\HeaderHelper;
+
+class TrustedHostsNetworkResolver implements MiddlewareInterface
```
samdark: Is it meant to be extended?

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: Yes, this is a special case where you can extends with custom function to recover the obfuscated host data. Also receive a `$ipDataList` variable to restore, which contains the valid elements of the chain. It is unknown to me if in such a case these input parameters are sufficient, I am even thinking about passing the `$request` variable.

Alternatively, it can be replaced with a callback.
```php
(new ...NetworkResolver(...))->withReverseObfuscate(function(...) {...})
```

However, this solution is more problematic at enterprise level:
```php
(new ...NetworkResolver(...))->withReverseObfuscate(function(...) {...})
vs
new MyNetworkResolver()
```

samdark: What's the use case for extending the class?

samdark:
https://github.com/yiisoft/yii2/pull/17676

tests/Middleware/TrustedHostsNetworkResolverTest.php
```diff
@@ -0,0 +1,191 @@
...
+                ['x-forwarded-for' => ['9.9.9.9', '5.5.5.5', '2.2.2.2']],
+                ['REMOTE_ADDR' => '127.0.0.1'],
+                [['hosts' => ['8.8.8.8', '127.0.0.1', '2.2.2.2']]],
+                '5.5.5.5',
+            ],
+            'forwardLevel1' => [
+                ['forward' => ['for=9.9.9.9', 'for=5.5.5.5', 'for=2.2.2.2']],
+                ['REMOTE_ADDR' => '127.0.0.1'],
+                [['hosts' => ['8.8.8.8', '127.0.0.1']]],
+                '2.2.2.2',
```
mikk150: if RFC literally says that it MUST BE appended always, why do we even test for addresses further than first one?

kamarton: https://tools.ietf.org/html/rfc7239#section-8.1

>  The "Forwarded" HTTP header field cannot be relied upon to be
   correct, as it may be modified, whether mistakenly or for malicious
   reasons, by every node on the way to the server, including the client
   making the request.

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: The `Forwarded` header supports obfuscated hosts, eg.
```
                            trusted           trusted             trusted           trusted
                               |                  |                 |                  |
           CLIENT <--------> PROXY <----------> PROXY2 <-------> PROXY3 <---------> SERVER
           1.1.1.1           2.2.2.2             --              3.3.3.3            4.4.4.4
Forwarded:                   for=1.1.1.1,    for=2.2.2.2,     for=_hidden 
```

PROXY2 is present in the network but for some reason is not public.

```php
class MyNetwrokResolver extends Trusted... {
  protected function reverseObfuscate(array $ipData, array $ipDataList): array {
     if($ipData['for'] === '_hidden' && /*before 2.2.2.2*/) {
       $ipData['ip'] = '2.2.2.0';  // any trusted IP
     }
     return parent::reverseObfuscate($ipData, $ipDataList)
  }
}
```

kamarton: This part is still under construction because it is not a good way to overwrite it with a fake IP address.

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: yes, for example

```php
class My... extends Trusted... {
  public function reverseObfuscate(..) ..
  public function isValidHost($host, $range, $validator): bool {
    if($range === ['any'] || $range !== ['*.yii.com']) {
      return parant::isValidHost(..);
    }
    // check reverse DNS or from IP list (https://www.cloudflare.com/ips-v4)
    $pattern = str_replace('\*','.*', preg_quote($range[0]));
    return preg_match("!^pattern$!", gethostbyaddr($host)) === 1;
  }
}

(new My...)->withAddedTrustedHosts(['8.8.8.8'], ...)
  ->withAddedTrustedHosts(['*.yii.com'], ...)
```

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: The final of inputs:
```php
    protected function reverseObfuscate(
        array $ipData,
        array $ipDataListValidated,
        array $ipDataListRemaining,
        RequestInterface $request
    ): array {

```

src/Helper/HeaderHelper.php
```diff
@@ -12,13 +12,30 @@ final class HeaderHelper
...
+    /**
+     * Explode header value to parameters (eg. q=2;version=6)
+     */
+    public static function getParameters(string $headerValue): array
     {
         $headerValue = trim($headerValue);
         if ($headerValue === '') {
             return [];
         }
         $parts = preg_split('/\s*;\s*/', $headerValue, -1, PREG_SPLIT_NO_EMPTY);
```
roxblnfk: I phink this parameters parser is incorrect.\
Please see [this tests](http://test.greenbytes.de/tech/tc2231/#attwithquotedsemicolon) for an example `Content-Disposition` header.
I assume the parsing of the `Content-Disposition: attachment; filename="Here's a semicolon;.html"` will be wrong

roxblnfk: Maybe we should create a new Issue for detailed testing of the `HeaderHelper` class

kamarton: https://github.com/yiisoft/yii-web/pull/183

kamarton:
@samdark reviewable (partially depend https://github.com/yiisoft/yii-web/pull/183)

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+            }
+        }
+        $untrustedHeaders = array_diff($trustedHeaders, $trustedHostData[self::DATA_KEY_TRUSTED_HEADERS] ?? []);
+        $request = $this->removeHeaders($request, $untrustedHeaders);
+        if ($trustedHostData === null) {
+            // No trusted host at all.
+            if ($this->notTrustedBranch !== null) {
+                return $this->notTrustedBranch->process($request, $handler);
+            }
+            $response = $this->responseFactory->createResponse(412);
```
kamarton: @samdark `403` or any?

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+     *
+     * This method can be extendable by overwriting eg. with reverse DNS verification.
+     */
+    protected function isValidHost(string $host, array $ranges, Ip $validator): bool
+    {
+        return $validator->ranges($ranges)->validate($host)->isValid();
+    }
+
+    /**
+     * Reverse obfuscating host data
```
samdark: It's not clear what "reverse obfuscation" stands for. Could you elaborate?

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+            if (!$request->hasHeader($protocolHeader)) {
+                continue;
+            }
+            if ($protocolHeader === $ipHeader && $ipListType === self::IP_HEADER_TYPE_RFC7239 && isset($ipData['protocol'])) {
+                $uri = $uri->withScheme($ipData['protocol']);
+                break;
+            }
+            $protocolHeaderValue = $request->getHeaderLine($protocolHeader);
+            foreach ($protocols as $protocol => $acceptedValues) {
+                if (\in_array($protocolHeaderValue, $acceptedValues)) {
```
samdark:
```suggestion
                if (\in_array($protocolHeaderValue, $acceptedValues, true)) {
```

src/Middleware/TrustedHostsNetworkResolver.php
samdark: I don't think it's a good idea to throw exception here. We're not trying to deny access based on IP, right? Setting `requestClientIp` to null sounds more appropriate.

kamarton: should I remove this part as well?
```php
            if ($this->notTrustedBranch !== null) {
                return $this->notTrustedBranch->process($request, $handler);
            }
```

samdark: If that's ability to set custom error handler then yes.

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+            self::DATA_KEY_TRUSTED_HEADERS => $trustedHeaders ?? self::DEFAULT_TRUSTED_HEADERS,
+            self::DATA_KEY_HOST_HEADERS => $hostHeaders,
+            self::DATA_KEY_URL_HEADERS => $urlHeaders,
+        ];
+        foreach ([
+                     self::DATA_KEY_HOSTS,
+                     self::DATA_KEY_TRUSTED_HEADERS,
+                     self::DATA_KEY_HOST_HEADERS,
+                     self::DATA_KEY_URL_HEADERS
+                 ] as $key) {
```
xepozz: Expose it to variable

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+    {
+        if ($attribute === '') {
+            throw new \RuntimeException('Attribute should not be empty');
+        }
+        $new = clone $this;
+        $new->attributeIps = $attribute;
+        return $new;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
xepozz: Please split this method on some methods.

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+    {
+        $actualHost = $request->getServerParams()['REMOTE_ADDR'];
+        $trustedHostData = null;
+        $trustedHeaders = [];
+        $ipValidator = $this->ipValidator ?? new Ip();
```
xepozz: May be it should be pass from constructor?

xepozz (changes requested):

kamarton:
@samdark how about if replaced `$hosts` with an `ip` validator?

```php
$nr->withAddedTrustedHosts(
  ['127.0.0.1', '8.8.8.8']
  /* ...  */);
// vs.
$nr->withAddedTrustedHosts(
  (new /* MyIp or */Ip())->dissallowIpv6()/* fastest check */->ranges(['127.0.0.1', '8.8.8.8'])
  /* ...  */);
```
with array:
- simple type
- require options for custom Ip validator
- require options for reverse obfuscating
- require options for custom validating case
- more checks for input

with `Ip` rule:
- more objects
- it is basically possible to use a custom Ip validator.
- less checks for input
- custom Ip validator allow to supports reverse obfuscating and custom validating case (extends of Ip class)

kamarton:
@samdark Yii2 did not support it, but many systems support `x-forwarded-port` (or alternate) headers. Is this make sense for us?
eg. https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-port

src/Middleware/TrustedHostsNetworkResolver.php
```diff
@@ -0,0 +1,510 @@
...
+
+    /**
+     * @return static
+     */
+    public function withIpValidator(Ip $ipValidator)
+    {
+        $new = clone $this;
+        $ipValidator = clone $ipValidator;
+        // force disable unacceptable validation
+        $new->ipValidator = $ipValidator->disallowSubnet()->disallowNegation();
```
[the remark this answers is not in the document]
kamarton: Cloning removed.

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: converted to variables

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: I see that this function of too many lines of code. It can be splitted, but for multiple functions, 4-5 parameters are transported, which is not very nice.

src/Middleware/TrustedHostsNetworkResolver.php
kamarton: more php doc added

samdark:
Merged. Thank you very much!

Issue #158, opened 2019-11-20
ErrorCatcher/HtmlRenderer: not migrated code in template


kamarton:
I think the template was moved from yii2. There are problems with `$this` referencing and private functions in code of template.
```
PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught Error: Using $this when not in object context in src/ErrorHandler/templates/exception.php:13
```

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

samdark:
The template is alright. I've fixed the pull request in question.

kamarton:
> The template is alright. I've fixed the pull request in question.

The template looks good but there is a bug somewhere, see #161.

samdark:
I'll check it.

Issue #155, opened 2019-11-19
ErrorCatcher: default content type depends on text/html renderer


kamarton:
### What steps will reproduce the problem?

I want to set a renderer for `text/html` and one for everything. Currently, I can only specify all other content types by overwriting text/html.

### What is the expected result?

I can specify a renderer separately for `text/html` and `*/*` at the same time.

### What do you get instead?

`text/html` and `*/*` are the same

### Suggests

Replace the following with `*/*` and add the HTML renderer to `*/*` in the list of renderers.
https://github.com/yiisoft/yii-web/blob/ff34903e0723201cc9962f1f3fbe0c8377bebe22/src/Middleware/ErrorCatcher.php#L113

https://github.com/yiisoft/yii-web/blob/ff34903e0723201cc9962f1f3fbe0c8377bebe22/src/Middleware/ErrorCatcher.php#L33

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

PR #159, opened 2019-11-20
[WIP] ErrorCatcher with default content type

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #155 

The #158 issue has a problem with the tests, which ruins a number of other tests. The `ErrorHandler` registers to error handler and causes problems in other tests. I don't know if this could cause a problem in the future.


samdark:
Fixed tests, formatting and merged. Thanks!

Issue #162, opened 2019-11-21
User identity stuck in RoadRunner environment


samdark:
Currently Identity instance is stored as state in https://github.com/yiisoft/yii-web/blob/master/src/User/User.php#L33 and it's not being cleaned up so it stucks in RoadRunner environment.

xepozz:
Should be moved to https://github.com/yiisoft/user

PR #163, opened 2019-11-21
Fix User identity stuck in RoadRunner environment

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | #162


samdark:
@rustamwin have you tested it locally?

rustamwin:
I haven't tested yet

rob006:
Wait, why user object is shared between two separate requests?

samdark:
@rob006 whole container is shared. That's how RoadRunner gains its performance. We can deal with it in three ways:

1. Avoid having state in services.
2. Clean up state individually for services that need it.
3. Have a special DI container mode for services that will be reset on each request.

samdark:
This change won't fix the problem for sure.

rob006:
@samdark Can we detect which services depends (directly on indirectly) on request? Basically all of them should be reset.

samdark:
@rob006 I don't think we can do that in runtime.

Issue #165, opened 2019-11-22
Emitter and streamed output


roxblnfk:
[We have a situation](https://github.com/yiisoft/yii-web/blob/22aaff78d9836f93561322bfacc3a4d852f0a0af/src/Emitter/SapiEmitter.php#L47) where the content of the stream are first completely unloaded into memory, and only then sent to the user.

The Emitter should output the stream in parts.

We should also consider supporting [HTTP Range Requests](https://tools.ietf.org/html/rfc7233#section-3).

kamarton:
It should also be possible to disable this feature, as it is mostly not needed.

roxblnfk:
I think that to send ranged Responses needed an individual Emitter, because for it to work correctly, we need to pass a Request object to it. And this is a different signature.

samdark:
Why do we need request in an emitter?

samdark:
Interface could be adjusted if needed.

kamarton:
I suggest middleware instead of modifying the Emitter interface.

roxblnfk:
> I suggest middleware instead of modifying the Emitter interface.

@smadark and i also came to this conclusion

PR #167, opened 2019-11-25
The Emitter now emit body in parts

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #165

### DIFF
--- a/src/Emitter/SapiEmitter.php
+++ b/src/Emitter/SapiEmitter.php
@@ -15,7 +15,7 @@ final class SapiEmitter implements EmitterInterface
 
     public function __construct(int $bufferSize = self::DEFAULT_BUFFER_SIZE)
     {
-        $this->bufferSize = $bufferSize > 0 ? $bufferSize : static::DEFAULT_BUFFER_SIZE;
+        $this->bufferSize = $bufferSize > 0 ? $bufferSize : self::DEFAULT_BUFFER_SIZE;
     }
 
     public function emit(ResponseInterface $response, bool $withoutBody = false): bool
--- a/src/Emitter/SapiEmitter.php
+++ b/src/Emitter/SapiEmitter.php
@@ -11,10 +11,11 @@ final class SapiEmitter implements EmitterInterface
 {
     private const NO_BODY_RESPONSE_CODES = [204, 205, 304];
     private $partSize;
+    private const DEFAULT_BUFFER_SIZE = 8388608; // 8MB
 
-    public function __construct(int $partSize = 8388608)
+    public function __construct(int $partSize = self::DEFAULT_BUFFER_SIZE)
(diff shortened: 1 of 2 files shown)


src/Emitter/SapiEmitter.php
```diff
@@ -10,6 +10,12 @@
 final class SapiEmitter implements EmitterInterface
 {
     private const NO_BODY_RESPONSE_CODES = [204, 205, 304];
+    private $partSize;
+
+    public function __construct(int $partSize = 8388608)
```
samdark: What is 8388608? Worth moving to a constant such as `private DEFAULT_BUFFER_SIZE` and adding phpdoc saying why it is exactly 8 megabytes.

samdark:
Merged. Thank you!

src/Emitter/SapiEmitter.php
```diff
@@ -10,6 +10,13 @@
 final class SapiEmitter implements EmitterInterface
 {
     private const NO_BODY_RESPONSE_CODES = [204, 205, 304];
+    private $bufferSize;
+    private const DEFAULT_BUFFER_SIZE = 8388608; // 8MB
+
+    public function __construct(int $bufferSize = self::DEFAULT_BUFFER_SIZE)
+    {
+        $this->bufferSize = $bufferSize > 0 ? $bufferSize : self::DEFAULT_BUFFER_SIZE;
```
kamarton: If zero or negative, the default approach is not really good. I suggests
- `?int .. = null` for parameter
- if negative or zero, throw InvalidParameterException
- if null then default.

src/Emitter/SapiEmitter.php
```diff
@@ -10,6 +10,13 @@
 final class SapiEmitter implements EmitterInterface
 {
     private const NO_BODY_RESPONSE_CODES = [204, 205, 304];
+    private $bufferSize;
+    private const DEFAULT_BUFFER_SIZE = 8388608; // 8MB
```
rustamwin: @samdark Do we have a clear order of properties? It looks a little weird

samdark: Oh, right. Constants first.

kamarton: Is it possible to create tests for this like changelog.md in yii2?

samdark: Yes, it is possible.

samdark: For now I've updated https://github.com/yiisoft/docs/blob/master/010-code-style.md#property-constant-and-method-placement to be clear about this case.

PR #168, opened 2019-11-25
Fix #93: Add non-verbose mode for ErrorHandler

### LINKED ISSUE #93: Ability for error handler to be switched to non-verbose mode
Useful for production where you don't want stack traces and details exposed.

samdark:
That would allow configuring it like the following:

```php
Yiisoft\Yii\Web\ErrorHandler\ErrorHandler::class => function (ContainerInterface $container) {
    $errorHandler = $container->get(ErrorHandler::class);
    return $errorHandler->withoutExposedDetails();
},
```

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #93

### DIFF
--- a/src/ErrorHandler/JsonRenderer.php
+++ b/src/ErrorHandler/JsonRenderer.php
@@ -9,6 +9,13 @@ final class JsonRenderer extends ThrowableRenderer
     public function render(\Throwable $t): string
     {
         return json_encode([
+            'message' => 'An internal server error occurred',
+        ], JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
+    }
+
+    public function renderVerbose(\Throwable $t): string
+    {
+        return json_encode([
             'type' => get_class($t),
             'message' => $t->getMessage(),
             'code' => $t->getCode(),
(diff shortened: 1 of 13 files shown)


src/ErrorHandler/JsonRenderer.php
```diff
@@ -7,6 +7,13 @@
 final class JsonRenderer extends ThrowableRenderer
 {
     public function render(\Throwable $t): string
+    {
+        return json_encode([
```
rustamwin: could we use `yiisoft/json`?

schmunk42:
Would be nice to have a use-case in the docs.

samdark:
@schmunk42 docs on error handler are to be done. Going to work on it soon.

src/ErrorHandler/JsonRenderer.php
samdark: Yes but I see no benefit of doing it in this particular case.

rustamwin (approved):

PR #169, opened 2019-11-26
SapiEmitter::emitBody() dupliicated code fix

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed PR  | https://github.com/yiisoft/yii-web/pull/167

The Emitter now emit body in parts by roxblnfk · Pull Request  167 · yiisoft yii-web

### DIFF
--- a/src/Emitter/SapiEmitter.php
+++ b/src/Emitter/SapiEmitter.php
@@ -64,20 +64,9 @@ final class SapiEmitter implements EmitterInterface
         if ($body->isSeekable()) {
             $body->rewind();
         }
-        $bytesToSend = $response->getBody()->getSize();
-        if ($bytesToSend > 0) {
-            while ($bytesToSend > 0 && !$body->eof()) {
-                $toRead = \min($this->bufferSize, $bytesToSend);
-                $data = $body->read($toRead);
-                echo $data;
-                $bytesToSend -= \strlen($data);
-                \flush();
-            }
-        } else {
-            while (!$body->eof()) {
-                echo $body->read($this->bufferSize);
-                \flush();
-            }
+        while (!$body->eof()) {
+            echo $body->read($this->bufferSize);
+            \flush();
         }
     }
 


roxblnfk:
First i wrote the code in such a way that in emitBody() could pass how many bytes to output. But after deleting this parameter, I did not notice that the code needs to be cleaned.
I sprinkle ashes on my head

samdark:
> Could such a case occur that eof () cannot reliably determine the end of the stream?

Unlikely.

samdark:
Merged. Thanks!

src/Emitter/SapiEmitter.php
```diff
@@ -64,20 +64,9 @@ private function emitBody(ResponseInterface $response): void
...
-                echo $data;
-                $bytesToSend -= \strlen($data);
-                \flush();
-            }
-        } else {
-            while (!$body->eof()) {
-                echo $body->read($this->bufferSize);
-                \flush();
-            }
+        while (!$body->eof()) {
```
roxblnfk: If we don’t look at what `-> getSize ()` returns to us, then maybe we should copy the behavior [from here](https://www.php.net/manual/en/streamwrapper.stream-eof.php)?

> When reading the whole file (for example, with file_get_contents()), PHP will call streamWrapper::stream_read() followed by streamWrapper::stream_eof() in a loop but as long as streamWrapper::stream_read() returns a non-empty string, the return value of streamWrapper::stream_eof() is ignored.

samdark: Why?

kamarton: https://www.php-fig.org/psr/psr-7/#34-psrhttpmessagestreaminterface

```
    /**
     * Returns true if the stream is at the end of the stream.
     *
     * @return bool
     */
    public function eof();
```

Expected operation is documented in the interface. If the stream is not implemented correctly, you need not worry.

roxblnfk: ok :-|

PR #175, opened 2019-11-26
SapiEmitter and tests update

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #167 #178 #179

SapiEmitterTest has been refactored. Now HTTP Headers can be tested without xdebug

Do you think it is necessary to write a test to verify that Emitter is emitting content in parts?


samdark:
> Do you think it is necessary to write a test to verify that Emitter is emitting content in parts?

Sounds like a good idea.

tests/Emitter/SapiEmitterTest.php
```diff
@@ -64,11 +145,11 @@ public function shouldNotOutputBodyIfEmitToldSo(): void
...
     {
         $length = 100;
-        $response = new Response(200, ['Content-length' => $length, 'X-Test' => 1], '');
+        $response = $this->createResponse(200, ['Content-Length' => $length, 'X-Test' => 1], '');
 
-        (new SapiEmitter())->emit($response);
+        $this->createEmitter()->emit($response);
 
-        $this->assertEquals(200, http_response_code());
+        $this->checkResponseCodeEquals(200);
```
kamarton: I suggests:
```php
$this->assertSame(200, $this->getResponseCode());
$this->assertSame(['x-test'...], $this->getResponseHeaders());
```

tests/Emitter/HTTPFunctions.php
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Remove previously set headers
+     */
+    public static function header_remove(?string $name = null): void
+    {
+        if ($name === null) {
+            self::$headers = [];
+        } else {
+            $name = strtolower($name);
+            if (key_exists($name, self::$headers)) {
```
kamarton: no condition is needed.
```php
unset(self::$headers[strtolower($name)]);
```

tests/Emitter/SapiEmitterTest.php
roxblnfk: `$this->assertSame(['x-test'...], $this->getResponseHeaders());`
This code does not take into account that the headers may be in different order

kamarton:
```php
$this->assertEquals(['a' => 'b', 'b' => 'a'], ['b' => 'a', 'a' => 'b']);  // this works fine
```

roxblnfk: > ```
> $this->assertEquals(['a' => 'b', 'b' => 'a'], ['b' => 'a', 'a' => 'b']);  // this works fine
> ```

`$this->assertEquals([1,2,3], [3,2,1]); // suicide`

samdark:
So behavior we need according to https://tools.ietf.org/html/rfc7230#section-3.3.2:

1. If there's Transfer-Encoding header, 1xx, 204, 304 response code or when request type is CONNECT (we may ignore this one) - don't send Content-Length.
2. The client should error when receiving invalid Content-Length according to RFC so either we can try correcting it or we can error with 500. I'm for erroring.
3. When Content-Length is not specified, send it automatically unless point 1.

tests/Emitter/SapiEmitterTest.php
```diff
@@ -1,24 +1,107 @@
...
+    {
+        HTTPFunctions::reset();
+    }
+
+    public function tearDown(): void
+    {
+        HTTPFunctions::reset();
+    }
+
+    public function testHTTPFunctions(): void
```
samdark: Maybe worth moving to a separate test. It's not testing SapiEmitter...

tests/Emitter/SapiEmitterTest.php
```diff
@@ -28,33 +111,31 @@ public function testEmit(): void
 
     /**
      * @test
-     */
+    //  */
```
samdark:
```suggestion
    */
```

tests/Emitter/SapiEmitterTest.php
samdark: How about multiple assertions with `assertContains()`?

tests/Emitter/SapiEmitterTest.php
```diff
@@ -82,22 +155,87 @@ public function contentLengthShouldNotBeOverwrittenIfPresent(): void
...
+        $this->assertEquals(['Content-Length: ' . strlen($body)], $this->getHeaders());
         $this->expectOutputString($body);
     }
 
     /**
-     * @param array $expected
+     * @test
      */
-    private function checkHeadersEquals($expected)
+    public function emitDuplicatedHeaders(): void
```
samdark:
```suggestion
    public function shouldEmitDuplicateHeaders(): void
```

samdark (changes requested):

src/Emitter/SapiEmitter.php
```diff
@@ -9,47 +9,56 @@
...
+            header_remove();
+            // send HTTP Status-Line
+            header(sprintf(
+                'HTTP/%s %d %s',
+                $response->getProtocolVersion(),
+                $status,
+                $response->getReasonPhrase()
+            ), true, $status);
+            // filter headers
+            $headers = $withoutContentLength
```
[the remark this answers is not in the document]
roxblnfk: You can't do that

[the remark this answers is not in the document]
roxblnfk: `$response->hasHeader('Content-Length')` used further

samdark (approved):

samdark:
Thank you!

PR #182, opened 2019-11-28
Fix #115: Adjust HTTP response code for invalid CSRF token

### LINKED ISSUE #115: Inappropriate HTTP 400 for CSRF token errors
Yii currently responds to requests that lack a valid CSRF token with a `BadRequestHTTPExeption`, which translates to HTTP error `400 Bad Request`:

https://github.com/yiisoft/yii2/blob/master/framework/web/Controller.php#L166

I suggest that this error code is semantically inappropriate. The relevant section of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.1) specifies that `400 Bad Request` is intended for:
> client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing)

I understand this to mean that `400 Bad Request` indicates something is wrong with the HTTP request itself rather than its payload, and it seems to me that CSRF validation failure does not constitute an error in the HTTP protocol; CSRF tokens aren't part of HTTP, and an HTTP request without a valid CSRF token is still perfectly valid as far as HTTP is concerned.

More appropriate responses might be:
* `422 Unprocessable entity` ([used by Rails](https://github.com/rails/rails/blob/fdd5894aa985651fb99e0ef1c2d3f3f604b2f659/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L17))
* `403 Forbidden`

[This discussion on Software Engineering Stack Exchange ](https://softwareengineering.stackexchange.com/q/329229) may also be of interest.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #115

### DIFF
--- a/tests/Middleware/CsrfTest.php
+++ b/tests/Middleware/CsrfTest.php
@@ -11,7 +11,6 @@ use PHPUnit\Framework\TestCase;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
-use TheSeer\Tokenizer\Token;
 use Yiisoft\Router\Method;
 use Yiisoft\Security\Random;
 use Yiisoft\Security\TokenMask;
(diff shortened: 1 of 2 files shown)


tests/Middleware/CsrfTest.php
```diff
@@ -79,31 +79,31 @@ public function getIsAlwaysAllowed()
...
     /**
      * @test
      */
-    public function emptyTokenInSessionResultIn400()
+    public function emptyTokenInSessionResultIn403()
     {
         $middleware = $this->createCsrfMiddlewareWithToken('');
         $response = $middleware->process($this->createPostServerRequestWithBodyToken($this->generateToken()), $this->createRequestHandler());
-        $this->assertEquals(400, $response->getStatusCode());
+        $this->assertEquals(403, $response->getStatusCode());
```
roxblnfk: Should we check that with an invalid token, the `_csrf` key is deleted from the session?

schmunk42:
I'd rather tend to error code `422`, see also https://github.com/yiisoft/yii-web/issues/115#issue-495235305

A `403` is more tied to a wrong login or auth-token, IMHO.

tests/Middleware/CsrfTest.php
samdark: I think that would be too much for this test.

roxblnfk (approved):

samdark:
Thanks for review, everyone.

PR #183, opened 2019-11-29
HeaderHelper: add getParameters(), fix getValuesAndParameters()

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed PR comment  | https://github.com/yiisoft/yii-web/pull/157#discussion_r351734462
| Related PR | https://github.com/yiisoft/yii-web/pull/157

### DIFF
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -17,16 +17,38 @@ final class HeaderHelper
         if ($headerValue === '') {
             return [];
         }
-        $parts = preg_split('/\s*;\s*/', $headerValue, -1, PREG_SPLIT_NO_EMPTY);
-        $output = [array_shift($parts)];
-        foreach ($parts as $part) {
-            [$key, $headerValue] = explode('=', $part, 2);
-            if (preg_match('/^"(?<value>.*)"$/', $headerValue, $matches) === 1) {
-                $headerValue = $matches['value'];
-            }
-            $output[$key] = $headerValue;
+        $parts = preg_split('/\s*;\s*/', $headerValue, 2, PREG_SPLIT_NO_EMPTY);
+        $output = [$parts[0]];
+        if (count($parts) === 1) {
+            return $output;
         }
-        return $output;
+        return array_merge($output, self::getParameters($parts[1]));
+    }
+
+    /**
+     * Explode header value to parameters (eg. q=2;version=6)
+     *
+     * @link https://tools.ietf.org/html/rfc7230#section-3.2.6
+     */
+    public static function getParameters(string $headerValue): array
+    {
+        $headerValue = trim($headerValue);
+        if ($headerValue === '') {
+            return [];
+        }
+        $output = new \stdClass();
+        $output->data = [];
+        do {
+            $headerValue = preg_replace_callback(
+                '/^\s*(?<parameter>\w+)\s*=\s*(?:(?<qoute>")(?<valueQuoted>[^"]+)\k<qoute>|(?<value>[!#$%&\'*+.^`|~\w\d-]+))\s*(?:;|$)/',
+                function ($matches) use ($output) {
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

src/Helper/HeaderHelper.php
```diff
@@ -17,16 +17,38 @@ public static function getValueAndParameters(string $headerValue): array
...
+     *
+     * @link https://tools.ietf.org/html/rfc7230#section-3.2.6
+     */
+    public static function getParameters(string $headerValue): array
+    {
+        $headerValue = trim($headerValue);
+        if ($headerValue === '') {
+            return [];
+        }
+        $output = new \stdClass();
```
xepozz: I think it should be simply ```$output = [];```

src/Helper/HeaderHelper.php
```diff
@@ -17,16 +17,38 @@ public static function getValueAndParameters(string $headerValue): array
...
+        $output->data = [];
+        do {
+            $headerValue = preg_replace_callback(
+                '/^\s*(?<parameter>\w+)\s*=\s*(?:"(?<valueQuoted>[^"]+)"|(?<value>[!#$%&\'*+.^`|~\w-]+))\s*(?:;|$)/',
```
roxblnfk: - Some parameter names can contain `*` (e.g. `filename*`)
- Quoted string can contain quoted-pair e.g. `\"`,  `\\` [[Basic Rules]](https://www.rfc-editor.org/rfc/rfc2616.html#section-2.2)
- Some headers can contain a listed values in parameters [[Accept]](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.1), [[Accept-Encoding]](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.3):
`Accept: audio/*; q=0.2, audio/basic`
`Accept-Encoding: gzip;q=1.0, identity; q=0.5, *;q=0`

I think it is impossible to solve this problem with one regular expression.
Moreover, some headlines require a personal touch.

src/Helper/HeaderHelper.php
samdark: https://github.com/yiisoft/yii-web/commit/8881b0421485031976912f5c4cc8321b8dde0468

src/Helper/HeaderHelper.php
samdark: @roxblnfk is it possible to pull-request a test for it? That would be much easier to fix then.

kamarton: > Some headers can contain a listed values in parameters [Accept], [Accept-Encoding]:

You mean that? https://github.com/yiisoft/yii-web/blob/8881b0421485031976912f5c4cc8321b8dde0468/src/Helper/HeaderHelper.php#L14

> Some parameter names can contain * (e.g. filename*)

in https://tools.ietf.org/html/rfc7230#section-3.2.6:
> Delimiters are chosen from the set of US-ASCII visual characters not allowed in a token (DQUOTE and "(),/:;<=>?@[\]{}").

`*` not allowed, but it's true that the token contains more characters.

> Quoted string can contain quoted-pair e.g. \", \\ [Basic Rules]
https://tools.ietf.org/html/rfc7230#section-3.2.6

agree

roxblnfk: > You mean that?

Yes sure :-| This is more about parsing a list of header values

kamarton: @roxblnfk this is a function that has been used a lot, would you to improve further with new tests in a new PR?

roxblnfk: @kamarton i have other tasks now. I can deal with this problem, but if only after some time

samdark: https://github.com/yiisoft/yii-web/issues/184

Issue #174, opened 2019-11-26
Response with attachment


kamarton:
- yii2: https://www.yiiframework.com/doc/api/2.0/yii-web-response#$downloadHeaders-detail
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

### I suggests

Implementing to `HttpHelper`.

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

samdark:
I'd suggest a separate class instead of putting it into a single large helper.

samdark:
See https://github.com/yiisoft/yii-web/blob/master/src/Cookie.php

kamarton:
https://github.com/yiisoft/yii-web/pull/185#issuecomment-560321613

kamarton:
```php
# RFC 2231/5987 Encoding: Character Sets
'2231iso' => ["filename=iso-8859-1''foo-%E4.html",null, null, ['filename' => 'foo-ä.html']],
'2231utf' => ["filename=UTF-8''foo-%c3%a4-%e2%82%ac.html",null, null, ['filename' => 'foo-ä-€.html']],
'2231noc' => ["filename=''foo-%c3%a4-%e2%82%ac.html", null,null, null, \InvalidArgumentException::class],

```

rustamwin:
What about implement in [yiisoft/http](https://github.com/yiisoft/http)?

sup-ham:
Maybe
- SendFile Middleware
- XSendFile Middleware

samdark:
@sup-ham how do you plan to use middleware there?

samdark:
https://github.com/yiisoft/http#contentdispositionheader-usage

PR #185, opened 2019-12-01
HeaderHelper::getParameters() fix #184

### LINKED ISSUE #184: Extra cases for HeaderHelper
See https://github.com/yiisoft/yii-web/pull/183#discussion_r352079796

> - Some parameter names can contain `*` (e.g. `filename*`)
> - Quoted string can contain quoted-pair e.g. `\"`,  `\\` [[Basic Rules]](https://www.rfc-editor.org/rfc/rfc2616.html#section-2.2)
> - Some headers can contain a listed values with parameters [[Accept]](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.1), [[Accept-Encoding]](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.3):
> `Accept: audio/*; q=0.2, audio/basic`
> `Accept-Encoding: gzip;q=1.0, identity; q=0.5, *;q=0`

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #184

@roxblnfk

### DIFF
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -71,7 +71,7 @@ final class HeaderHelper
         $output = [];
         do {
             $headerValue = preg_replace_callback(
-                '/^(?<parameter>' . self::PATTERN_ATTRIBUTE . ')=(?<value>' . self::PATTERN_VALUE . ')(?:;|$)/',
+                '/^[ \t]*(?<parameter>' . self::PATTERN_ATTRIBUTE . ')[ \t]*=[ \t]*(?<value>' . self::PATTERN_VALUE . ')[ \t]*(?:;|$)/',
                 static function ($matches) use (&$output, $lowerCaseParameter, $lowerCaseValue) {
                     $value = $matches['value'];
                     if (substr($matches['value'], 0, 1) === '"') {
--- a/src/Helper/HeaderHelper.php
+++ b/src/Helper/HeaderHelper.php
@@ -40,18 +40,18 @@ final class HeaderHelper
      * @param string $headerValue
      * @return array first element is the value, and key-value are the parameters
      */
-    public static function getValueAndParameters(string $headerValue): array
+    public static function getValueAndParameters(string $headerValue, bool $lowerCaseValue = true, bool $lowerCaseParameter = true, bool $lowerCaseParameterValue = true): array
     {
         $headerValue = trim($headerValue);
         if ($headerValue === '') {
             return [];
         }
         $parts = explode(';', $headerValue, 2);
-        $output = [$parts[0]];
+        $output = [$lowerCaseValue ? strtolower($parts[0]) : $parts[0]];
         if (count($parts) === 1) {
             return $output;
         }
-        return array_merge($output, self::getParameters($parts[1]));
+        return $output + self::getParameters($parts[1], $lowerCaseParameter, $lowerCaseParameterValue);
     }
 
     /**
@@ -59,7 +59,7 @@ final class HeaderHelper
      *
      * @link https://tools.ietf.org/html/rfc7230#section-3.2.6
      */
-    public static function getParameters(string $headerValue): array
+    public static function getParameters(string $headerValue, bool $lowerCaseParameter = true, $lowerCaseValue = true): array
     {
         $headerValue = trim($headerValue);
         if ($headerValue === '') {
@@ -72,13 +72,13 @@ final class HeaderHelper
         do {
             $headerValue = preg_replace_callback(
                 '/^(?<parameter>' . self::PATTERN_ATTRIBUTE . ')=(?<value>' . self::PATTERN_VALUE . ')(?:;|$)/',
-                static function ($matches) use (&$output) {
+                static function ($matches) use (&$output, $lowerCaseParameter, $lowerCaseValue) {
                     $value = $matches['value'];
                     if (substr($matches['value'], 0, 1) === '"') {
                         // unescape + remove first and last quote
                         $value = preg_replace('/\\\\(.)/', '$1', substr($value, 1, -1));
                     }
-                    $output[$matches['parameter']] = $value;
+                    $output[$lowerCaseParameter ? strtolower($matches['parameter']) : $matches['parameter']] = $lowerCaseValue ? strtolower($value) : $value;
                 }, $headerValue, 1, $count);
             if ($count !== 1) {
                 throw new \InvalidArgumentException('Invalid input: ' . $headerValue);
@@ -92,8 +92,9 @@ final class HeaderHelper
      * @param string|string[] $values Header value as a comma-separated string or already exploded string array.
      * @see getValueAndParameters
      * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+     * @link https://www.ietf.org/rfc/rfc2045.html#section-2
      */
-    public static function getSortedValueAndParameters($values): array
+    public static function getSortedValueAndParameters($values, bool $lowerCaseValue = true, bool $loweCaseParameter = true, bool $lowerCaseParameterValue = true): array
     {
         if (is_string($values)) {
             $values = preg_split('/\s*,\s*/', trim($values), -1, PREG_SPLIT_NO_EMPTY);
@@ -106,14 +107,16 @@ final class HeaderHelper
         }
         $output = [];
         foreach ($values as $value) {
-            $parse = self::getValueAndParameters($value);
-            $q = $parse['q'] ?? 1.0;
+            $parse = self::getValueAndParameters($value, $lowerCaseValue, $loweCaseParameter, $lowerCaseParameterValue);
+            // case-insensitive "q" parameter
+            $q = $parse['q'] ?? $parse['Q'] ?? 1.0;
 
             // min 0.000 max 1.000, max 3 digits, without digits allowed
             if (is_string($q) && preg_match('/^(?:0(?:\.\d{1,3})?|1(?:\.0{1,3})?)$/', $q) === 0) {
                 throw new \InvalidArgumentException('Invalid q factor');
             }
             $parse['q'] = (float)$q;
+            unset($parse['Q']);
             $output[] = $parse;
         }
         usort($output, static function ($a, $b) {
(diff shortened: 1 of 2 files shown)


samdark (approved):

rustamwin (approved):

src/Helper/HeaderHelper.php
```diff
@@ -6,8 +6,37 @@
...
+     */
+    private const PATTERN_VALUE = '(?:' . self::PATTERN_QUOTED_STRING . '|' . self::PATTERN_TOKEN . ')';
+
     /**
      * Explode header value to value and parameters (eg. text/html;q=2;version=6)
+     *
+     * @link https://www.rfc-editor.org/rfc/rfc2616.html#section-3.6
+     * transfer-extension      = token *( ";" parameter )
      * @param string $headerValue
      * @return array first element is the value, and key-value are the parameters
```
roxblnfk: What do you think about those cases when the values of the header are specified by a comma separated list?
Should this function return a list of arrays?
`Accept: audio/*; q=0.2, audio/basic`
```
[
  ['audio/*', 'q' => '0.2'],
  ['audio/basic'],
]
```
If so, can it be better to pass a `Request` object instead `$headerValue` parameter here with the name of the header, or an array of strings received with `$request->getHeader($headerName)`?

If we get only one value, we need to write that the list is not processed and we get only the first value

roxblnfk:
I would like to draw your attention to [rfc2231](https://tools.ietf.org/html/rfc2231)
Do we need to support that?

roxblnfk:
https://github.com/kamarton/yii-web/pull/1

samdark:
@roxblnfk, @kamarton ready for final review?

kamarton:
> @roxblnfk, @kamarton ready for final review?

no.

tests/Helper/HeaderHelperTest.php
```diff
@@ -134,37 +150,75 @@ public function testSortedAcceptTypesFromRequest(RequestInterface $request, arra
...
+            'invalidQuotes2' => ['a="test',null, null, null, \InvalidArgumentException::class],
+            'invalidQuotes3' => ['a=test"',null, null, null, \InvalidArgumentException::class],
+            'invalidQuotes4' => ['a=te"st',null, null, null, \InvalidArgumentException::class],
+            'invalidEmptyValue' => ['a=b; c=',null, null, null, \InvalidArgumentException::class],
+            'invalidEmptyParam' => ['a=b; ;c=d',null, null, null, \InvalidArgumentException::class],
+            'semicolonAtEnd' => ['a=b;',null, null, null, \InvalidArgumentException::class],
+            'comma' => ['a=test,test',null, null, null, \InvalidArgumentException::class],
+
+            # true syntax
+            // 'spaces2' => [' a = b ; c = "d" ',null, null, ['a' => 'b', 'c' => 'd']],
```
kamarton: I can't find an RFC that specifically this case in general, but as I see it, spaces are not allowed on many points in [RFC2616](https://www.rfc-editor.org/rfc/rfc2616.html).

tests/Helper/HeaderHelperTest.php
```diff
@@ -134,37 +150,75 @@ public function testSortedAcceptTypesFromRequest(RequestInterface $request, arra
...
+            'case' => ['A=TEST;TEST=B',true, false, ['a' => 'TEST', 'test' => 'B']],
+            'case2' => ['A=TEST;TEST=B',false, false, ['A' => 'TEST', 'TEST' => 'B']],
+            'case3' => ['A=TEST;TEST=B',false, true, ['A' => 'test', 'TEST' => 'b']],
+            // 'spaces1' => ['a=b; c="d" ',null, null, ['a' => 'b', 'c' => 'd']],
+            'spaces3' => ['a=b c',null, null, null, \InvalidArgumentException::class],
+            'percent' => ['a=%1;b="foo-%32-bar"',null, null, ['a' => '%1', 'b' => 'foo-%32-bar']],
+
+            # Invalid syntax but most browsers take a first parameter
```
kamarton: This is an interesting case, and I would rather support an exception. Invalid use causes quite a few problems, eg. why not accept the first or why not the last.

I think it is likely that the yii framework will mainly receive request by standard browsers. In the case of an API server, this can also force the client (and developer) site to send valid requests.

tests/Helper/HeaderHelperTest.php
```diff
@@ -134,37 +150,75 @@ public function testSortedAcceptTypesFromRequest(RequestInterface $request, arra
...
+            'brokenToken' => ['a=foo[1](2).html',null, null, null, \InvalidArgumentException::class],
+
+            'brokenSyntax1' => ['a==b',null, null, null, \InvalidArgumentException::class],
+            'brokenSyntax2' => ['a *=b',null, null, null, \InvalidArgumentException::class],
+            # Invalid syntax but most browsers accept the umlaut with warn
+            'brokenToken2' => ['a=foo-ä.html',null, null, ['a' => 'foo-ä.html']],
+            # Invalid syntax but most browsers accept the umlaut with warn
+            'brokenToken3' => ['a=foo-Ã¤.html',null, null, ['a' => 'foo-Ã¤.html']],
+
+            # RFC 2231/5987 Encoding: Character Sets
```
kamarton: I think this usually only appears in response. #174

With low priority, I can imagine a separate issue.

src/Helper/HeaderHelper.php
```diff
@@ -55,8 +92,9 @@ static function ($matches) use (&$output) {
      * @param string|string[] $values Header value as a comma-separated string or already exploded string array.
      * @see getValueAndParameters
      * @link https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+     * @link https://www.ietf.org/rfc/rfc2045.html#section-2
      */
-    public static function getSortedValueAndParameters($values): array
+    public static function getSortedValueAndParameters($values, bool $lowerCaseValue = true, bool $loweCaseParameter = true, bool $lowerCaseParameterValue = true): array
```
xepozz:
```suggestion
    public static function getSortedValueAndParameters($values, bool $lowerCaseValue = true, bool $lowerCaseParameter = true, bool $lowerCaseParameterValue = true): array
```

src/Helper/HeaderHelper.php
```diff
@@ -69,14 +107,16 @@ public static function getSortedValueAndParameters($values): array
         }
         $output = [];
         foreach ($values as $value) {
-            $parse = self::getValueAndParameters($value);
-            $q = $parse['q'] ?? 1.0;
+            $parse = self::getValueAndParameters($value, $lowerCaseValue, $loweCaseParameter, $lowerCaseParameterValue);
```
xepozz:
```suggestion
            $parse = self::getValueAndParameters($value, $lowerCaseValue, $lowerCaseParameter, $lowerCaseParameterValue);
```

xepozz (changes requested):
Too much functions with default value or nullable in arguments. Can you redesign they that they disappear?
Default value or nullable arguments are bad smells

kamarton:
> Too much functions with default value or nullable in arguments. Can you redesign they that they disappear?
> Default value or nullable arguments are bad smells

```php
getSortedValueAndParameters($values, bool $lowerCaseValue = true, bool $lowerCaseParameter = true, bool $lowerCaseParameterValue = true): array
```

how about config paramter?

```php
getSortedValueAndParameters($values, $config = self::LOWER_CASE_VALUE | self::LOWER_CASE_PARAMETER_NAME | self::LOWER_CASE_PARAMETER_VALUE): array
```

xepozz:
Why not 2 or many undependent functions that do their work?

And some improvements to HeaderHelper.
1. Non-static methods.
2. Collection for headers.
    It would be nice to work with headers as a collection of classes.
    For example like this:
    `$header->getValue()`, `$header->getName()`, `$header->getName()->equalsTo('CaseInSenSiTiViTyName')`
    Or something this:
    ```php
    $rawHeaders = ['Language' => 'en-US'];
    $headersCollection = new HeadersCollection($rawHeaders);
    $helper = new HeaderHelper();
    $headers = $headersCollection->asArray();
    // $headers = ['Language' => 'en-US'];
    $headers = $helper->toLowerCase($headersCollection)->asArray();
    // $headers = ['language' => 'en-US'];
    ```
    Or other way.

@kamarton WDYT?

tests/Helper/HeaderHelperTest.php
roxblnfk: I offered only test cases. On controversial moments, when the syntax is broken, we are free to do as we prefer: to be more strict or to do as most browsers

tests/Helper/HeaderHelperTest.php
roxblnfk: I also think that it is not practical to implement it now

tests/Helper/HeaderHelperTest.php
samdark: It is better to check https://tools.ietf.org/html/rfc7230#section-3.2, 2616 is obsolete and less definitive.

```
 header-field   = field-name ":" OWS field-value OWS

     field-name     = token
     field-value    = *( field-content / obs-fold )
     field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
     field-vchar    = VCHAR / obs-text

     obs-fold       = CRLF 1*( SP / HTAB )
                    ; obsolete line folding
                    ; see Section 3.2.4

    obs-text       = %x80-FF

     OWS            = *( SP / HTAB )
               ; optional whitespace
```

tests/Helper/HeaderHelperTest.php
samdark: If we aren't going to implement it it is better to remove commented out tests as well.

kamarton: removed and saved https://github.com/yiisoft/yii-web/issues/174#issuecomment-568657384

tests/Helper/HeaderHelperTest.php
kamarton: modified, optional spaces are allowed

tests/Helper/HeaderHelperTest.php
kamarton:
```php
            # Invalid syntax but most browsers take a first parameter
            // 'sameName' => ['a=T1;a="T2"',null, null, ['a' => 'T1']],
            // 'sameNameCase' => ['aa=T1;Aa="T2"',null, null, ['aa' => 'T1']],
```
@samdark first, last or throw?

roxblnfk: >first, last or throw?

I think most browsers take the first value because they are trying to take the maximum from the Headers until the RFC is violated.
Declaring the second parameter of the same name is the violation, so all that was normal before, is accepted (first declaration).

kamarton: first is the winner.

samdark:
Thanks!

Issue #190, opened 2019-12-06
Typed property Yiisoft\Yii\Web\User\User::$identity must not be accessed before initialization


thenotsoft:
Typed property Yiisoft\Yii\Web\User\User::$identity must not be accessed before initialization
when call Yiisoft\Yii\Web\User\User::isGuest()

| Q                | A
| ---------------- | ---
| Version          | 1.0.?
| PHP version      |  7.4
| Operating system |

thenotsoft:
fixed #191

rob006:
How can you get this error, if this is not typed property (yet)?

thenotsoft:
If you do not use typing, this error is absent, since the minimum version in the composer.json is now 7.4, I set the typing locally for the test and received an error.

rob006:
So you just broken this class locally, a then reported this as a bug?...

PR #191, opened 2019-12-07
adjustment for php 7.4

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #190

### DIFF
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -17,31 +17,12 @@ class User
     private const SESSION_AUTH_EXPIRE = '__auth_expire';
     private const SESSION_AUTH_ABSOLUTE_EXPIRE = '__auth_absolute_expire';
 
-    /**
-     * @var IdentityRepositoryInterface
-     */
-    private $identityRepository;
-
-    /**
-     * @var AccessCheckerInterface
-     */
-    private $accessChecker;
-
-    /**
-     * @var IdentityInterface
-     */
-    private $identity;
-
-    /**
-     * @var Session
-     */
-    private $session;
-
-    /**
-     * @var EventDispatcherInterface
-     */
-    private $eventDispatcher;
-
+    private IdentityRepositoryInterface $identityRepository;
+    private EventDispatcherInterface $eventDispatcher;
+    private ?AccessCheckerInterface $accessChecker = null;
+    private ?IdentityInterface $identity = null;
+    private ?Session $session = null;
+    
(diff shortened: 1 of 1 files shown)


src/User/User.php
```diff
@@ -17,31 +17,12 @@ class User
...
-    /**
-     * @var EventDispatcherInterface
-     */
-    private $eventDispatcher;
-
+    private IdentityRepositoryInterface $identityRepository;
+    private EventDispatcherInterface $eventDispatcher;
+    private ?AccessCheckerInterface $accessChecker = null;
+    private ?IdentityInterface $identity = null;
+    private ?Session $session = null;
```
xepozz:
```suggestion
    private ?Session $session;
```
By default all properties have `null` value.

src/User/User.php
```diff
@@ -17,31 +17,12 @@ class User
...
-
-    /**
-     * @var EventDispatcherInterface
-     */
-    private $eventDispatcher;
-
+    private IdentityRepositoryInterface $identityRepository;
+    private EventDispatcherInterface $eventDispatcher;
+    private ?AccessCheckerInterface $accessChecker = null;
+    private ?IdentityInterface $identity = null;
```
xepozz:
```suggestion
    private ?IdentityInterface $identity;
```

src/User/User.php
```diff
@@ -17,31 +17,12 @@ class User
...
-    private $session;
-
-    /**
-     * @var EventDispatcherInterface
-     */
-    private $eventDispatcher;
-
+    private IdentityRepositoryInterface $identityRepository;
+    private EventDispatcherInterface $eventDispatcher;
+    private ?AccessCheckerInterface $accessChecker = null;
```
xepozz:
```suggestion
    private ?AccessCheckerInterface $accessChecker;
```

src/User/User.php
thenotsoft: no, in 7.4 if you use typing you need to initialize the property before it is called
if used like this it will throw an error when calling a property
`Fatal error: Uncaught Error: Typed property User::$session
must not be accessed before initialization`
now if the property can contain null you need to set it to null by default it is not set

xepozz: I didn't know that. Thank you

samdark:
Doesn't seem to fix anything yet. We'll add types later if needed. But thanks for effort anyway.

PR #192, opened 2019-12-09
Fix User::isGuest() return user is not a guest

thenotsoft:
If call User::isGuest() before getIdentity() or setIdentity this always returns the user is not a guest since identity is null and not an instance of GuestIdentity

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -205,7 +205,7 @@ class User
      */
     public function isGuest(): bool
     {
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

PR #194, opened 2019-12-16
Subfolder support

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | https://github.com/yiisoft/yii-demo/issues/11 https://github.com/yiisoft/yii-demo/pull/34

Added SubfolderMiddleware which supports routing when webroot is not the same folder as public


sup-ham:
somewhat less consistent, but only this middleware has `Middleware` suffix in its name
image

Issue #173, opened 2019-11-26
HTTP cache


kamarton:
- yii2: https://www.yiiframework.com/doc/guide/2.0/en/caching-http
- [`cache-control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
- [`etag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)
- [Caching in HTTP (RFC2616#13)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html)

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

PR #199, opened 2020-01-01
Porting http cache

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #173


src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,282 @@
...
+     * @var string the name of the cache limiter to be set when [session_cache_limiter()](https://secure.php.net/manual/en/function.session-cache-limiter.php)
+     * is called. The default value is an empty string, meaning turning off automatic sending of cache headers entirely.
+     * You may set this property to be `public`, `private`, `private_no_expire`, and `nocache`.
+     * Please refer to [session_cache_limiter()](https://secure.php.net/manual/en/function.session-cache-limiter.php)
+     * for detailed explanation of these values.
+     *
+     * If this property is `null`, then `session_cache_limiter()` will not be called. As a result,
+     * PHP will send headers according to the `session.cache_limiter` PHP ini setting.
+     */
+    private ?string $sessionCacheLimiter = '';
```
samdark: I don't think we can use `session_cache_limiter()` or `header_*` because headers should be written into PSR-7 object and then sent using https://github.com/yiisoft/yii-web/blob/master/src/Emitter/SapiEmitter.php.

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * @var bool a value indicating whether this filter should be enabled.
+     */
+    private bool $enabled = true;
```
samdark: That's not needed. If you don't want it to be enabled, remove it from middleware stack.

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,282 @@
...
+            $lastModified = call_user_func($this->lastModified, $request, $this->params);
+        }
+        if ($this->etagSeed !== null) {
+            $seed = call_user_func($this->etagSeed, $request, $this->params);
+            if ($seed !== null) {
+                $etag = $this->generateEtag($seed);
+            }
+        }
+
+        $this->sendCacheControlHeader($request);
```
samdark: How is it used?

samdark (changes requested):

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,282 @@
...
+
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Log\LoggerInterface;
+use Yiisoft\Router\Method;
+use Yiisoft\Yii\Web\Session\SessionInterface;
+use function in_array;
```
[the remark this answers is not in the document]
rustamwin: PHPStorm EA :smile:

src/Middleware/HttpCache.php
rustamwin: You are right :)

src/Middleware/HttpCache.php
rustamwin: hmm..

src/Middleware/HttpCache.php
```diff
@@ -107,15 +86,10 @@ public function __construct(
...
-        if (!$this->enabled) {
-            return $handler->handle($request);
-        }
-
         $method = $request->getMethod();
         if (
-            !in_array($method, [Method::GET, Method::HEAD])
-            || $this->lastModified === null
-            && $this->etagSeed === null
+            !\in_array($method, [Method::GET, Method::HEAD])
```
samdark: Seems we're using it wrong :( `Method` is part of router package.

rustamwin: It's used in some middleware

samdark: Yes and it's wrong...

rustamwin: Then should we use strings instead?

samdark: These constants are useful so I think we can move Method from router package to web and use strings in the router package instead.

kamarton: how about `yiisoft/http-utilities` package with `Method` class and `HeaderHelper` class

rustamwin: @kamarton Did you mean the new `yiisoft/http-utilities` package?

kamarton: yes

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,197 @@
...
+    {
+        $this->weakEtag = $weakTag;
+    }
+
+    public function setParams($params): void
+    {
+        $this->params = $params;
+    }
+
+    public function setCacheControlHeader(?string $header): void
```
kamarton: A builder class would be practical. For example:
```php
..->setCacheControlHeader((string) (new CacheControlHeaderBuilder())
    ->public()
    ->maxAge(300)
    ->sharedMaxAge(60)
  );
```

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,197 @@
...
+     * @see http://tools.ietf.org/html/rfc7232#section-2.2
+     */
+    private $lastModified;
+
+    /**
+     * @var callable a PHP callback that generates the ETag seed string.
+     * The callback's signature should be:
+     *
+     * ```php
+     * function ($request, $params)
```
kamarton: `function(ServerRequestInterface $request, $params): ?string`

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,197 @@
...
+final class HttpCache implements MiddlewareInterface
+{
+    private const DEFAULT_HEADER = 'public, max-age=3600';
+
+    /**
+     * @var callable a PHP callback that returns the UNIX timestamp of the last modification time.
+     * The callback's signature should be:
+     *
+     * ```php
+     * function ($request, $params)
```
kamarton: `function(ServerRequestInterface $request, $params): int`

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * @var mixed additional parameters that should be passed to the [[lastModified]] and [[etagSeed]] callbacks.
+     */
+    private $params;
+
+    /**
+     * @var string the value of the `Cache-Control` HTTP header. If null, the header will not be sent.
+     * @see http://tools.ietf.org/html/rfc2616#section-14.9
+     */
+    private ?string $cacheControlHeader = self::DEFAULT_HEADER;
```
kamarton: Does it make sense to use this middleware without a header?

src/Middleware/HttpCache.php
```diff
@@ -0,0 +1,197 @@
...
+    {
+        $this->responseFactory = $responseFactory;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $method = $request->getMethod();
+        if (
+            !\in_array($method, [Method::GET, Method::HEAD])
+            || ($this->lastModified === null && $this->etagSeed === null)
```
kamarton: In my opinion, these are mandatory data (`lastModified` and `etagSeed`). If not configured, it can be left out of the middleware stack.

src/Middleware/HttpCache.php
rustamwin: > how about yiisoft/http-utilities package with Method class and HeaderHelper class?

@samdark ?

samdark: https://github.com/yiisoft/http. Other packages code is already updated.

src/Middleware/HttpCache.php
samdark: Yes. See https://tools.ietf.org/html/rfc7232#section-2.3

tests/Middleware/HttpCacheTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Http\Method;
+use Yiisoft\Yii\Web\Middleware\HttpCache;
+
+class HttpCacheTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function validCacheResult(): void
```
samdark: What's exactly tested here? 200 should be returned if content was modified since X along with the body. The X is current time. It's a bit confusing overall.

tests/Middleware/HttpCacheTest.php
```diff
@@ -0,0 +1,68 @@
...
+            'If-Modified-Since' => $time,
+        ];
+        $response = $middleware->process($this->createServerRequest(Method::GET, $headers), $this->createRequestHandler());
+        $this->assertEquals(200, $response->getStatusCode());
+    }
+
+    /**
+     * @test
+     */
+    public function invalidCacheResult(): void
```
samdark: 304 is not invalid cache. It's, in fact, valid i.e. result was not modified.

In this case:

1. Body should be empty.
2. Last-Modified header should be present.

tests/Middleware/HttpCacheTest.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * @test
+     */
+    public function validCacheResult(): void
+    {
+        $time = \time();
+        $middleware = $this->createMiddlewareWithLastModified($time);
+        $headers = [
+            'If-Modified-Since' => $time,
```
samdark: Shouldn't it be formatted as `gmdate('D, d M Y H:i:s', $time) . 'GMT',`?

tests/Middleware/HttpCacheTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Nyholm\Psr7\Response;
+use Nyholm\Psr7\ServerRequest;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Http\Method;
+use Yiisoft\Yii\Web\Middleware\HttpCache;
+
+class HttpCacheTest extends TestCase
```
samdark: Additional test ideas:

1. `If-Modified-Since` should be ignored if `If-None-Match` header present.
2. Whole set of tests for ETags.
3. Test that middleware only works for GET/HEAD requests. For other request methods it should do nothing.

samdark (changes requested):

src/Middleware/HttpCache.php
rustamwin: @samdark

tests/Middleware/HttpCacheTest.php
rustamwin: Yes, I will add these tests later...

src/Middleware/HttpCache.php
rustamwin: I don't understand. What did you mean?

src/Middleware/HttpCache.php
rustamwin: I think the header builder can be implemented in [yiisoft/http](https://github.com/yiisoft/http)

tests/Middleware/HttpCacheTest.php
rustamwin: Tests done

xepozz:
Can you avoid annotation `@test`? Why not use name of method that starts with `test*`?

samdark:
@xepozz why avoiding it? It's fine to use it.

kamarton:
> @xepozz why avoiding it? It's fine to use it.

In my opinion, it is better to have a uniform code. Other tests use the prefix `test`.

xepozz:
> @xepozz why avoiding it? It's fine to use it.

1. Annotation takes a lot of space (3 redundant lines)
3. Method without prefix `test` may interfere when you are looking some method with using autocomplete (for example type `$this->` then press `ctrl` + `space` in PHPStorm)
3. A lot of test functions already declared with using prefix `test`. As @kamarton said we should be follow a uniform code.

src/Middleware/HttpCache.php
samdark: One of these is mandatory, not both. So the current code is OK.

samdark (approved):

samdark:
Merged. Thank you!

PR #202, opened 2020-01-04
Bugfix user permission

Abdujabbor:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | bug on User::can method replaced illegal method name of interface


samdark:
Thanks!

Issue #63, opened 2018-02-23
The suggestion to Yii2 core team is to design RateLimiter filter good enough not only for API calls, but to make it more reusable in simple controllers.


gauravpatel1110:
Here is the use case:

    Limit number of API calls globally - 5 requests/second
    Limit number of API calls for action1 - 30 requests/minute
    Limit number of API calls for action2 - 60 requests/minute

    Limit number of successfull user registrations (action /register) - 1/hour
    Limit number of failed user logins (action /login) - 5 requests/15 minutes
    Add possibility to have 1 shared limit for several actions

  Make This With FileCache.

samdark:
Rate limiter should be implemented as middleware.

kamarton:
I would recommend separate the actual counting unit and using it anywhere in the code, eg sending SMS limit.

samdark:
This one is now done.

PR #203, opened 2020-01-05
Rate limiter implementation

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #63


src/Middleware/RateLimiter.php
```diff
@@ -0,0 +1,149 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * Rate limiter limits the number of requests that could be made within a certain period of time
+ */
+final class RateLimiter implements MiddlewareInterface
```
samdark: It it still "leaky bucket" algorithm as it was in Yii 2?

src/Middleware/RateLimiter.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public function setCacheKey(string $key): self
+    {
+        $this->cacheKey = $key;
+
+        return $this;
+    }
+
+    public function setCacheKeyByCallback(callable $callback): self
```
samdark:
```suggestion
    public function setCacheKeyCallback(callable $callback): self
```

src/Middleware/RateLimiter.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $this->cacheKeyCallback = $callback;
+
+        return $this;
+    }
+
+    public function setCacheTtl(int $ttl): self
```
samdark: Should we use more abstract name like "period" or "interval"?

src/Middleware/RateLimiter.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $this->cacheTtl = $ttl;
+
+        return $this;
+    }
+
+    public function setAutoIncrement(bool $increment): self
```
samdark: What's the idea behind it? Why would one disable increment? Also, there's no way to increment manually except writing to cache (and that's not a good idea).

src/Middleware/RateLimiter.php
```diff
@@ -0,0 +1,149 @@
...
+        $value++;
+
+        $this->setCounterValue($value);
+    }
+
+    private function setupCacheParams(ServerRequestInterface $request): void
+    {
+        $this->cacheKey = $this->setupCacheKey($request);
+
+        if (!$this->hasCounterValue()) {
```
samdark: There seems to be no need for setting value to 0 since 0 is default returned if there's nothing in cache.

tests/Middleware/RateLimiterTest.php
```diff
@@ -0,0 +1,161 @@
...
+        $this->assertEquals(200, $response->getStatusCode());
+    }
+
+    /**
+     * @test
+     */
+    public function moreThanDefaultNumberOfRequestsIsNotAllowed(): void
+    {
+        $cache = $this->getCache();
+        $this->setRateLimiterCurrentRequestNumber($cache, 1000);
```
samdark: While that is efficient, the test is knowing about cache key names. It might be worth it to call process N times instead. That would be closer to what happens in reality when requests are processed.

samdark (changes requested):

src/Middleware/RateLimiter.php
samdark: Looks like a huge simplification code-wise 👍

src/Middleware/RateLimiter.php
romkatsu: Yes, that's right. Cleaned.

src/Middleware/RateLimiter.php
romkatsu: renamed to interval.

src/Middleware/RateLimiter.php
romkatsu: The idea was that you could use this implementation not only to limit requests, but also for example-SMS. 

In this case, you will have to write to the cache manually. 

In my opinion, the obligation of middleware is still to check the limits. 

There was an idea to take out work with a cache in a separate class and there to implement the increment function, but I'm not sure that this is a good idea.


In case of SMS:

```
$middleware = (new RateLimiter($cache, $responseFactory))
    ->setLimit(100)
    ->setInterval(86400)
    ->setCacheKey('send-sms-users');
```

```
 Route::get('/')
      ->prepend($middleware)
      ->to(new ActionCaller(SiteController::class, 'index', $container))
      ->name('site/sms')
```

```
public function sms(CacheInterface $cache, SmsService $smsService): ResponseInterface
{
    if (!$this->user->isGuest()) {
        $smsService->send($this->user);
        $cache->set('send-sms-users', 100);
    }
    ...
}
```

tests/Middleware/RateLimiterTest.php
romkatsu: fixed

src/Middleware/RateLimiter.php
romkatsu: Option with a separate class:

RateLimiter:
```
final class RateLimiter implements MiddlewareInterface
{
    private int $limit = 1000;

    private bool $autoincrement = true;

    private RateLimiterCounter $counter;

    private ResponseFactoryInterface $responseFactory;

    public function __construct(RateLimiterCounter $counter, ResponseFactoryInterface $responseFactory)
    {
        $this->counter = $counter;
        $this->responseFactory = $responseFactory;
    }

    {
        $this->counter->init($request);

        if (!$this->isAllowed()) {
            return $this->createErrorResponse();
        }

        if ($this->autoincrement) {
            $this->counter->increment();
        }

        return $handler->handle($request);
    }

    public function setLimit(int $limit): self
    {
        $this->limit = $limit;

        return $this;
    }

    public function setAutoIncrement(bool $increment): self
    {
        $this->autoincrement = $increment;

        return $this;
    }

    private function createErrorResponse(): ResponseInterface
    {
        $response = $this->responseFactory->createResponse(429);
        $response->getBody()->write('Too Many Requests');

        return $response;
    }

    private function isAllowed(): bool
    {
        return $this->counter->getCounterValue() < $this->limit;
    }
}
```

RateLimiterCounter:
```
final class RateLimiterCounter
{
    private int $ttl = 360;

    private ?string $key = null;

    /**
     * @var callable
     */
    private $keyCallback;

    private CacheInterface $storage;

    public function __construct(CacheInterface $cache)
    {
        $this->storage = $cache;
    }

    public function init(ServerRequestInterface $request): self
    {
        $this->key = $this->generateKey($request);

        return $this;
    }

    public function setKeyCallback(callable $callback): self
    {
        $this->keyCallback = $callback;

        return $this;
    }

    public function setKey(string $key): self
    {
        $this->key = $key;

        return $this;
    }

    public function setTtl(int $ttl): self
    {
        $this->ttl = $ttl;

        return $this;
    }

    public function increment(): void
    {
        $value = $this->getCounterValue();
        $value++;

        $this->setCounterValue($value);
    }

    public function getCounterValue(): int
    {
        $this->checkOptions();
        return $this->storage->get($this->key, 0);
    }

    private function setCounterValue(int $value): void
    {
        $this->checkOptions();
        $this->storage->set($this->key, $value, $this->ttl);
    }

    private function generateKey(ServerRequestInterface $request): string
    {
        if ($this->keyCallback !== null) {
            return \call_user_func($this->keyCallback, $request);
        }

        return $this->key ?? $this->generateKeyFromRequest($request);
    }

    private function generateKeyFromRequest(ServerRequestInterface $request): string
    {
        return strtolower('rate-limiter-' . $request->getMethod() . '-' . $request->getUri()->getPath());
    }

    private function checkOptions(): void
    {
        if ($this->key === null) {
            throw new \RuntimeException('The cache key cannot be null');
        }
    }
}
```


Usage:
```
        $counter = (new RateLimiterCounter($cache))
           ->setKey('sms-send');

        $middleware = (new RateLimiter($counter, $responseFactory))
            ->setLimit(10)
            ->setAutoIncrement(false);
```

manual increment:
```
        (new RateLimiterCounter($cache))
            ->setKey('sms-send')
            ->increment();
```

rustamwin (changes requested):
Without a database repository implementation?

romkatsu:
> Without a database repository implementation?

I didn't plan on it, but I can try. What tasks can this be useful for?

rustamwin:
> I didn't plan on it, but I can try. What tasks can this be useful for?

See https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimiter.php and https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimitInterface.php

romkatsu:
> > I didn't plan on it, but I can try. What tasks can this be useful for?
> 
> See https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimiter.php and https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimitInterface.php

added storage implementation, now it will be possible to add storage in the database

src/Middleware/RateLimiter.php
romkatsu: added counter implementation

src/RateLimiter/CounterInterface.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\RateLimiter;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+interface CounterInterface
```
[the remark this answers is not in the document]
romkatsu: UserLoginCounter or CommentsPerPostCounter?

I didn't think it would hurt anyone. 

Remove it?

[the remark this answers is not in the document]
samdark: Would the implementation be different for UserLoginCounter or CommentsPerPostCounter?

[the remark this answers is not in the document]
romkatsu: hardly. removed

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return $this;
+    }
+
+    public function increment(): void
+    {
+        $value = $this->getCounterValue();
+        $value++;
+
+        $this->storage->set($this->id, $value, $this->interval);
```
samdark: While it makes code simple, this rate limiter implementation (fixed window) results in possibility of traffic spikes on the edge of the interval. The behavior for interval = 360 (6 minutes) and limit = 10000 is that after 10000 requests there will be 504 HTTP response for exactly 6 minutes. Then there could be quick 10000 requests again and another 6 minute pause etc. So you'll still have traffic spikes this way. That's a no go for a good rate limiter.

Good algorithms preventing traffic spikes are:

1. Leaky bucket https://en.wikipedia.org/wiki/Leaky_bucket. That one is in Yii 2.
2. Sliding window.

See:

1. https://blog.cloudflare.com/counting-things-a-lot-of-different-things/
2. https://habr.com/ru/post/448438/

samdark: I'd implement GCRA:

- https://brandur.org/rate-limiting
- https://blog.ian.stapletoncordas.co/2018/12/understanding-generic-cell-rate-limiting.html

kamarton: And counts are lost. for example
1. A request get counter value (X)
2. B request get counter value (X)
3. A requrest set counter value (X+1)
4. B request set counter value (X+1)

In fact, X + 2. In multi server environments, this can mean a lot of lost counting. 
I recommend that the storage interface have a `incrementAndGet()` method. [Redis](https://redis.io/commands/incr), mysql allow counting in atomic.

samdark: Yes, increment operation should be atomic.

rustamwin (changes requested):
All setters should be `void` or `with` prefixed
```php
public function withLimit(int $limit): self
{
    $new = clone $this
    $new->limit = $limit;
    return $new;
}
```

src/RateLimiter/CacheCounterStorage.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\RateLimiter;
+
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * CacheCounterStorage stores counter value in cache
+ */
+final class CacheCounterStorage implements CounterStorageInterface
```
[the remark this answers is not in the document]
samdark: The purpose of it is to store counter data. How would you name it?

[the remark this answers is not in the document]
kamarton: It's not about storage, it's about counting.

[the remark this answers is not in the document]
samdark: Nope. https://github.com/yiisoft/yii-web/pull/203/files#diff-e94ff4fb5bda21fe7b12afa594018f4c is about counting. This one is about storage. If these should be separate is another question.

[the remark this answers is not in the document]
kamarton: Why storage? There is no need to read it, it is always only increment, if the result is higher than the limit then it is too many requests

```php
CacheCounter implements CounterInterface // or CounterServiceInterface
```

src/RateLimiter/CounterStorageInterface.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\RateLimiter;
+
+/**
+ * Counter value storage
+ */
+interface CounterStorageInterface
```
kamarton: I suggest simplify it

```php
interface CounterInterface
{
  public function incrementAndGet(string $id, int $interval): int;
}
```

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($this->hasIdCallback()) {
+            return \call_user_func($this->idCallback, $request);
+        }
+
+        return $this->id ?? $this->generateIdFromRequest($request);
+    }
+
+    private function generateIdFromRequest(ServerRequestInterface $request): string
+    {
+        return strtolower('rate-limiter-' . $request->getMethod() . '-' . $request->getUri()->getPath());
```
kamarton: generate is not public, the `strtolower()` can be omitted.

src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,75 @@
...
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * RateLimiter limits the number of consequential requests ({@see RateLimiter::$limit}) that could be processed per
+ * {@see Counter::$interval}. If the number is exceeded, middleware responds with HTTP code 429, "Too Many Requests"
+ * until limit expires.
+ */
+final class RateLimiter implements MiddlewareInterface
+{
+    private int $limit = 1000;
```
kamarton: Does the default value make sense?

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,102 @@
...
+    public function __construct(CounterStorageInterface $storage)
+    {
+        $this->storage = $storage;
+    }
+
+    public function init(ServerRequestInterface $request): self
+    {
+        $this->id = $this->generateId($request);
+
+        if (!$this->storage->has($this->id)) {
```
kamarton: This is already the task of `CounterInterface`.

src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,75 @@
...
+
+    private Counter $counter;
+
+    private ResponseFactoryInterface $responseFactory;
+
+    private bool $autoincrement = true;
```
kamarton: I don't really understand the point of leaving auto-increment. If autoincrement is disabled, it is more like just a "checker".

src/RateLimiter/CacheCounterStorage.php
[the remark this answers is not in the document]
romkatsu: if CounterStorageInterface renamed to CounterInterface and CacheCounterStorage renamed to CacheCounter. How will be named [Counter](https://github.com/yiisoft/yii-web/blob/4d22d84680b3984164ba8adfd3042328002ec59e/src/RateLimiter/Counter.php) ?

src/RateLimiter/Counter.php
romkatsu: Query request parameters can be anything, I just want the ID to be the same in terms of case.

src/RateLimiter/CacheCounterStorage.php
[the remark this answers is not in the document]
romkatsu: if CounterStorageInterface renamed to CounterInterface and CacheCounterStorage renamed to CacheCounter. How will be named Counter ?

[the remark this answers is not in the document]
rustamwin: I think the [Counter](https://github.com/yiisoft/yii-web/blob/4d22d84680b3984164ba8adfd3042328002ec59e/src/RateLimiter/Counter.php) class is not needed. Each `counter` should be individually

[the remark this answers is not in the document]
kamarton: `Counter` in current implementation is useless. Currently, a single `Counter` can be specified for middleware -> can be implemented to middleware.

kamarton:
Current implementation not allow multiple limit descriptor in a single middleware. It is necessary to record more intervals and more limits, eg.
```
ID                    interval  limit
sms/user/12             60s     2
sms/user/12           3600s     10
sms/users               60s     40
sms/users             3600s     250

```

src/RateLimiter/CacheCounterStorage.php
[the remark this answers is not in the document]
romkatsu: > I think the [Counter](https://github.com/yiisoft/yii-web/blob/4d22d84680b3984164ba8adfd3042328002ec59e/src/RateLimiter/Counter.php) class is not needed. Each `counter` should be individually

we've already [discussed this](https://github.com/yiisoft/yii-web/pull/203#discussion_r363946997).

What are possible examples?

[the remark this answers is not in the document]
romkatsu: > `Counter` in current implementation is useless. Currently, a single `Counter` can be specified for middleware -> can be implemented to middleware.

Why useless?  You can also use multiple instances of middleware.

```
$counter1 = (new Counter($storage))->setInterval(60);
$counter2 = (new Counter($storage))->setInterval(86400);

$middleware1 = (new RateLimiter($counter1, $responseFactory))
    ->withLimit(10);

$middleware2 = (new RateLimiter($counter2, $responseFactory))
    ->withLimit(100);

Route::get('/')
    ->prepend($middleware1)
    ->to(new ActionCaller(SiteController::class, 'index', $container))
    ->name('site/index')

Route::get('/')
   ->prepend($middleware2)
   ->to(new ActionCaller(SiteController::class, 'index', $container))
   ->name('site/test')
```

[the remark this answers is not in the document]
kamarton: > Why useless? You can also use multiple instances of middleware.

the current implementation is the same as:

```php
$counter = new CacheCounter();
$middleware1 = (new RateLimiter($counter, $responseFactory))
    ->withLimit(10)
    ->withInterval(60);
$middleware1 = (new RateLimiter($counter, $responseFactory))
    ->withLimit(100)
    ->withInterval(86400);
```

romkatsu:
Created a [new pull request](https://github.com/yiisoft/yii-web/pull/204) with the implementation of the GCRA.

I'm closing this one.

PR #204, opened 2020-01-10
Rate limiter implementation(gcra)

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #63 

previous pull request #203 

[GCRA](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm)

https://brandur.org/rate-limiting
https://blog.ian.stapletoncordas.co/2018/12/understanding-generic-cell-rate-limiting.html


src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,83 @@
...
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * RateLimiter limits the number of consequential requests ({@see CacheCounter::$limit}) that could be processed per
+ * {@see CacheCounter::$interval}. If the number is reached, middleware responds with HTTP code 429, "Too Many Requests"
+ * until limit expires.
+ */
+final class RateLimiter implements MiddlewareInterface
+{
+    private CacheCounter $counter;
```
kamarton: designed for cache interface only?

rustamwin:
It would be great if the logic was ported from Yii2

samdark:
@rustamwin GCRA is better about resource consumption than Yii 2 leaky bucket.

src/RateLimiter/RateLimiter.php
romkatsu: Yes. Do you think you need other storage?

kamarton: The cache (or storage) interface does not support atomically operation.

kamarton:
GCRA is good for transactional system, but simple atomic counter design is low cost operation.
- GCRA require transaction (if need atomcy), GET operation and SET operation (depend on condition) - 2 operations. 
- The atomic counter works with one operation (increment+get), and most cache services support the operation.

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private function setStorageValue(float $theoreticalArrivalTime): void
+    {
+        $this->storage->set($this->id, $theoreticalArrivalTime);
+    }
+
+    private function getArrivalTime(): int
+    {
+        return time() * self::MILLISECONDS_PER_SECOND;
```
kamarton: Will this solution support 1000 / 10sec limits?

src/RateLimiter/RateLimiter.php
samdark: Yes, indeed, its own interface would fit it much better since without atomic operation of incrementing a counter it won't work well under high loads. There will be request spikes.

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,99 @@
...
+        return false;
+    }
+
+    private function checkParams(): void
+    {
+        if ($this->id === null) {
+            throw new \RuntimeException('The counter id not set');
+        }
+
+        if ($this->limit < 1) {
```
kamarton: move it to constructor

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,99 @@
...
+            throw new \InvalidArgumentException('The limit must be a positive value.');
+        }
+
+        if ($this->period < 1) {
```
kamarton: move it to constructor

kamarton (changes requested):

src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->counter->setId($this->generateId($request));
+
+        if ($this->counter->limitIsReached()) {
+            return $this->createErrorResponse();
+        }
+
+        return $handler->handle($request);
+    }
+
+    public function withCounterIdCallback(callable $callback): self
```
kamarton: return with a new instance (use clone) or rename to `counterIdCallback()`

src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $this->counterIdCallback = $callback;
+
+        return $this;
+    }
+
+    public function withCounterId(string $id): self
```
kamarton: return with a new instance (use clone) or rename to `counterId()`

src/RateLimiter/RateLimiter.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $this->counterId = $id;
+
+        return $this;
+    }
+
+    private function createErrorResponse(): ResponseInterface
+    {
+        $response = $this->responseFactory->createResponse(429);
```
kamarton: add response headers. 
- https://tools.ietf.org/html/draft-polli-ratelimit-headers-01
- https://github.com/yiisoft/yii2/blob/da1fddb57617f6659cc4572db35454d366eff562/framework/filters/RateLimiter.php#L140-L142

src/RateLimiter/RateLimiter.php
kamarton: The parameter can also be set to NULL. `?callable $callback`

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,99 @@
...
+    private function remainingEmpty(float $theoreticalArrivalTime): bool
+    {
+        $allowAt = $theoreticalArrivalTime - $this->period;
+
+        return ((floor($this->arrivalTime - $allowAt) / $this->getEmissionInterval()) + 0.5) < 1;
+    }
+
+    private function getStorageValue(): float
+    {
+        return $this->storage->get($this->id, (float)$this->arrivalTime);
```
kamarton: add prefix to $id. The current implementation may conflict with other cache usage.

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private function setStorageValue(float $theoreticalArrivalTime): void
+    {
+        $this->storage->set($this->id, $theoreticalArrivalTime);
```
kamarton: add prefix to $id. The current implementation may conflict with other cache usage.

src/RateLimiter/CacheCounter.php
romkatsu: fixed

src/RateLimiter/CacheCounter.php
romkatsu: fixed

src/RateLimiter/CacheCounter.php
romkatsu: Yes, supports

src/RateLimiter/CacheCounter.php
romkatsu: fixed

src/RateLimiter/CacheCounter.php
romkatsu: fixed

src/RateLimiter/RateLimiter.php
romkatsu: fixed

src/RateLimiter/RateLimiter.php
romkatsu: fixed

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->period = $period * self::MILLISECONDS_PER_SECOND;
+        $this->storage = $storage;
+    }
+
+    public function setId(string $id): void
+    {
+        $this->id = self::ID_PREFIX . $id;
+    }
+
+    public function limitIsReached(): bool
```
kamarton: The name `limitIsReached` is not good because it not only checks.

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        if ($this->id === null) {
+            throw new \RuntimeException('The counter id not set');
```
kamarton: what do you think about using [`\LogicException`](https://www.php.net/manual/en/class.logicexception.php)?

romkatsu: fixed

src/RateLimiter/CacheCounter.php
romkatsu: rename to incrementAndGetResult

src/RateLimiter/RateLimiter.php
romkatsu: done

src/RateLimiter/RateLimiterMiddleware.php
```diff
@@ -0,0 +1,96 @@
...
+
+        return $this->counterId ?? $this->generateIdFromRequest($request);
+    }
+
+    private function generateIdFromRequest(ServerRequestInterface $request): string
+    {
+        return strtolower($request->getMethod() . '-' . $request->getUri()->getPath());
+    }
+
+    public function addHeaders(ResponseInterface $response, RateLimitResult $result): ResponseInterface
```
samdark: private?

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function setId(string $id): void
+    {
+        $this->id = self::ID_PREFIX . $id;
+    }
+
+    public function incrementAndGetResult(): RateLimitResult
+    {
+        if ($this->id === null) {
+            throw new \LogicException('The counter id not set');
```
samdark: It looks like setting `id` is mandatory. If so, it worth moving it to constructor.

samdark:
https://github.com/yiisoft/yii-web/runs/387629182 is interesting. I've changed no code. Then it fixed by itself. Looks like algorithm issue.

src/RateLimiter/CacheCounter.php
```diff
@@ -0,0 +1,103 @@
...
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\RateLimiter;
+
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * CacheCounter implements generic сell rate limit algorithm https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm
+ */
+final class CacheCounter
```
samdark: I think that there are two things user may want to customize:

- Where/how to store counter (storage).
- How to calculate it (algorithm).

Worth thinking about making it flexible in these places.

src/RateLimiter/CacheCounter.php
romkatsu: fixed

src/RateLimiter/RateLimiterMiddleware.php
romkatsu: fixed

samdark:
Seems tests are still shaky. Failing from time to time...

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,104 @@
...
+    private function calculateTheoreticalArrivalTime(float $theoreticalArrivalTime): float
+    {
+        return max($this->arrivalTime, $theoreticalArrivalTime) + $this->emissionInterval;
+    }
+
+    private function calculateRemaining(float $theoreticalArrivalTime): int
+    {
+        $allowAt = $theoreticalArrivalTime - $this->period;
+
+        return (int)((floor($this->arrivalTime - $allowAt) / $this->emissionInterval) + 0.5);
```
samdark: Why do we need 0.5 here?

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * @param float $theoreticalNextIncrementTime
+     * @return int the number of remaining requests in the current time period
+     */
+    private function calculateRemaining(float $theoreticalNextIncrementTime): int
+    {
+        $incrementAllowedAt = $theoreticalNextIncrementTime - $this->period;
+
+        return (int)((floor($this->lastIncrementTime - $incrementAllowedAt) / $this->incrementInterval) + 0.5);
```
samdark: Why do we need 0.5 here?

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,128 @@
...
+
+    private int $limit;
+
+    /**
+     * @var float maximum interval before next increment, in milliseconds
+     * In GCRA it is known as emission interval.
+     */
+    private float $incrementInterval;
+
+    private ?string $id = null;
```
xepozz: Why it nullable? Can we pass value from constructor instead setter?

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Counter implements generic сell rate limit algorithm (GCRA) that ensures that after reaching the limit futher
+ * increments are distributed equally.
+ *
+ * @link https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm
+ */
+final class Counter implements CounterInterface
+{
+    public const ID_PREFIX = 'rate-limiter-';
```
xepozz: Maybe for debuggin and tests add the method `getCacheKey()`  and set this constant `private`?
Later it may used in any services as `$key = $this->counter->getCacheKey()` instead of `$key = Counter::ID_PREFIX . ?`

src/RateLimiter/Counter.php
kamarton: https://github.com/yiisoft/yii-web/pull/204/files#diff-b529e2c30fdfe6601a69f5e3f7f63dafR83

if null then generated from request automatically

samdark: 1. To have it in the interface.
2. To allow it to be filled automatically by middleware if it's not set.

src/RateLimiter/Counter.php
samdark: Done.

samdark:
Another test is planned to ensure that distribution of increments is OK after reaching the limit. @romkatsu is on it.

tests/RateLimiter/CounterTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function statisticsShouldBeCorrectWhenLimitIsNotReached(): void
+    {
+        $counter = new Counter(2, 5, new ArrayCache());
+        $counter->setId('key');
+
+        $statistics = $counter->incrementAndGetResult();
+        $this->assertEquals(2, $statistics->getLimit());
+        $this->assertEquals(1, $statistics->getRemaining());
+        $this->assertEquals(2500, $statistics->getReset());
+        $this->assertFalse($statistics->isLimitReached());
```
kamarton: add more loop
```php
$statistics = $counter->incrementAndGetResult();
$this->assertEquals(2, $statistics->getLimit());
$this->assertEquals(1, $statistics->getRemaining());   // <--- remaining = 1
$this->assertEquals(2500, $statistics->getReset());
$this->assertFalse($statistics->isLimitReached());
// next loop
$statistics = $counter->incrementAndGetResult();
// ...
```

src/RateLimiter/RateLimiterMiddleware.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        $new = clone $this;
+        $new->counterId = $id;
+
+        return $new;
+    }
+
+    private function createErrorResponse(): ResponseInterface
+    {
+        $response = $this->responseFactory->createResponse(429);
```
kamarton: https://tools.ietf.org/html/rfc6585#section-4: add `retry-after` header

src/RateLimiter/CounterStatistics.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function getRemaining(): int
+    {
+        return $this->remaining;
+    }
+
+    /**
+     * @return int milliseconds to wait until the rate limit resets
+     */
+    public function getReset(): int
```
kamarton: I suggests 
- `@return float seconds ...` `float` with second unit, OR
- use absolute timestamp: It is more accurate because the sending of the response causes a delay. In this case too, float is a better choice if seconds are not enough.

samdark: Seems common practice is to use seconds: https://stackoverflow.com/questions/16022624/examples-of-http-api-rate-limiting-http-response-headers. Not sure if millisecond precision is needed.

samdark: Absolute timestamp isn't common practice.

samdark: @romkatsu what do you think?

src/RateLimiter/RateLimiterMiddleware.php
samdark: Good idea.

tests/RateLimiter/CounterTest.php
samdark: 0 is when limit is reached. It's out of the scope of this test.

tests/RateLimiter/CounterTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Yiisoft\Yii\Web\Tests\RateLimiter;
+
+use InvalidArgumentException;
+use PHPUnit\Framework\TestCase;
+use Yiisoft\Cache\ArrayCache;
+use Yiisoft\Yii\Web\RateLimiter\Counter;
+
+final class CounterTest extends TestCase
```
kamarton: Add another test where a full limit runs through.
```php
        $counter = new Counter(2, 5, new ArrayCache());
        $counter->setId('key');
$statistics = $counter->incrementAndGetResult();
$this->assertEquals(2, $statistics->getLimit());
$this->assertEquals(1, $statistics->getRemaining());   // <--- remaining = 1
$this->assertEquals(2500, $statistics->getReset());
$this->assertFalse($statistics->isLimitReached());
// next loop
$statistics = $counter->incrementAndGetResult();
// ...
```

src/RateLimiter/CounterStatistics.php
kamarton: This class is not dedicated to HTTP headers. When you actually set the headers, you can calculate the absolute, relative, and any other mode.+

kamarton:
```php
    {
        $this->counter->setId($this->generateId($request));
        $result = $this->counter->incrementAndGetResult();  // <--- result.reset = 3 seconds
        if ($result->isLimitReached()) {
            $response = $this->createErrorResponse();
        } else {
            $response = $handler->handle($request);      // <-------- it runs in 10 seconds
        }
        return $this->addHeaders($response, $result);     // <---- The header will be 3 seconds, but the window expiration time is already 0 seconds.
    }
```

src/RateLimiter/RateLimiterMiddleware.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $this->counter->setId($this->generateId($request));
+        $result = $this->counter->incrementAndGetResult();
+
+        if ($result->isLimitReached()) {
+            $response = $this->createErrorResponse();
+        } else {
+            $response = $handler->handle($request);
```
kamarton: If this section runs for a long time, the value of `$ result` will be outdated.

src/RateLimiter/CounterStatistics.php
samdark: @kamarton makes sense.

src/RateLimiter/CounterInterface.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\RateLimiter;
+
+/**
+ * CounterInterface describes rate limiter counter
+ */
+interface CounterInterface
```
kamarton: I suggests

```php
interface CounterInterface
{
  public function id(string $id): self;
  // required for all algorithm of counters (GCRA, counter based...) to gnerate statistic
  public function limit(int $limit): self;    
  // required for all algorithm of counters (GCRA, counter based...) to gnerate statistic
  public function interval(int $interval): self;
  public function incrementAndGetStatistic(): CounterStatistic;
  // required for real statistic after the one or more middleware is handled
  public function getStatistic(): CounterStatistic;
}
```

src/RateLimiter/CounterInterface.php
```diff
@@ -0,0 +1,23 @@
...
+     * @param string $id set counter ID
+     * Counters with non-equal IDs are counted separately.
+     */
+    public function setId(string $id): void;
+
+    /**
+     * Counts one request as done and returns result containing statistics and rate limiting decision
+     * @return CounterStatistics
+     */
+    public function incrementAndGetResult(): CounterStatistics;
```
xepozz: This method should be divided into 2 methods: `increment` and `getResult`. 
See [CQRS](wikipedia.org/wiki/CQRS)

xepozz (changes requested):

src/RateLimiter/CounterInterface.php
kamarton: if divided, atomic operation is not possible.

xepozz: which one?

src/RateLimiter/RateLimiterMiddleware.php
rustamwin: https://github.com/yiisoft/http
```suggestion
        $response = $this->responseFactory->createResponse(Status::TOO_MANY_REQUESTS);
```

tests/RateLimiter/CounterTest.php
romkatsu: done

src/RateLimiter/RateLimiterMiddleware.php
romkatsu: @rustamwin fixed

src/RateLimiter/CounterStatistics.php
romkatsu: fixed. now returns the time

src/RateLimiter/RateLimiterMiddleware.php
romkatsu: Method getReset now returns the time.

src/RateLimiter/CounterInterface.php
samdark: I don't think expansion of the interface is needed.

1. limit and interval are counter configuration details. These aren't used by middleware or any client so it's better to leave these as constructor arguments.
2. There's currently no client for `getStatistic()`. What's the use case?

kamarton:
```php
    {
        $this->counter->setId($this->generateId($request));
        $result = $this->counter->incrementAndGetResult();
        if ($result->isLimitReached()) {
            $response = $this->createErrorResponse();
        } else {
            $response = $handler->handle($request);  // <--- if handle is run long time, then $result is totally outdated
        }
        return $this->addHeaders($response, $result);
    }
```

If handle is run long time, then $result is totally outdated, therefore, a method is needed that get the current state. (`getResult === getStatistic`)

rate limiter - Google Táblázatok

samdark: Then incrementAndGetStatistic() result isn't used. Do you propose to separate increment and getting result?

kamarton: if limit is reached, then result is usable, because in that case he immediately returns. I do not think it can be separated, because in that case increment loses the possibility of atomic operation.

kamarton: If a separated, then in this case it is easy for both requests to reach the limit, even though one could have run.

rate limiter - Google Táblázatok

src/RateLimiter/CounterInterface.php
kamarton: `incrementAndGetResult`

src/RateLimiter/CounterInterface.php
samdark: So you want this?

```php
{
    $this->counter->setId($this->generateId($request));
    $result = $this->counter->incrementAndGetResult();

    if ($result->isLimitReached()) {
        $response = $this->createErrorResponse();
    } else {
        $result = $this->counter->getResult();
    }

    return $this->addHeaders($response, $result);
}
```

kamarton: Yes

src/RateLimiter/Counter.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private function getLastStoredTheoreticalNextIncrementTime(): float
+    {
+        return $this->storage->get($this->getCacheKey(), (float)$this->lastIncrementTime);
+    }
+
+    private function storeTheoreticalNextIncrementTime(float $theoreticalNextIncrementTime): void
+    {
+        $this->storage->set($this->getCacheKey(), $theoreticalNextIncrementTime);
```
kamarton: Set ttl to preserve leak.
`Ttl=$theoreticalNextincrementTime - time() +1`

kamarton (changes requested):

src/RateLimiter/CounterInterface.php
samdark: In this case we'll have to:

1. Store last increment time.
2. Do two cache reads instead of one in case of not rate-limited request.

src/RateLimiter/Counter.php
samdark: What do you mean by "to preserve leak"?

src/RateLimiter/CounterInterface.php
samdark: I'm not sure it worth it. Yes, headers may get outdated.

In the worst case client will be mis-informed about rate allowance and will make extra request that would be responded with "429 Too Many Requests". I don't see it as a big deal.

Same is likely to happen because of networking reasons: if internet connection is slow or shaky.

src/RateLimiter/CounterInterface.php
samdark: That's not the case to apply CQRS to. There may be a valid reason to have two methods as described by https://github.com/yiisoft/yii-web/pull/204#discussion_r366701262 though. Let's discuss there.

src/RateLimiter/CounterInterface.php
kamarton: > Do two cache reads instead of one in case of not rate-limited request.

depend on implementation.

> In the worst case client will be mis-informed about rate allowance and will make extra request that will be responded with "429 Too Many Requests". I don't see it as a big deal.

I don't think the interface should be designed less functionality. If you feel it is not necessary, you may want to design the middleware so that it is not capable.

src/RateLimiter/Counter.php
kamarton: Creates a cache record for each `$id` that never expires.

If the `$id` has many variant, the cache will easily be full of 5-10 thousand or more records that will never be deleted. However, in all likelihood, these records are never needed at the same time.

src/RateLimiter/CounterInterface.php
kamarton: > limit and interval are counter configuration details. These aren't used by middleware or any client so it's better to leave these as constructor arguments.

The current design is not logical. The resulting objects have minimal reusability.

current implementation:
```php
$counter1 = new CacheCounter(10,100, $cache);
$middleware1 = new RateLimiter($counter1, ...);
$counter2 = new CacheCounter(100,100, $cache);
$middleware2 = new RateLimiter($counter2, ...);
```

instead of
```php
$counter=new Counter($cache);
$middleware1= new RateLimiter($counter, 10, 100, ...);
$middleware2= new RateLimiter($counter, 100, 100, ...);
// or extended implementacion
$middlewareMulti = new RateLimiter($counter, ...)
->withAddLimit(10,100)
->withAddLimit(5000, 3600);
```

src/RateLimiter/Counter.php
samdark: Right. Thanks.

samdark: Done.

src/RateLimiter/CounterInterface.php
samdark: > depend on implementation.

You have to do a second read if you want fresh information. There's no way you can avoid that.

> I don't think the interface should be designed less functionality. If you feel it is not necessary, you may want to design the middleware so that it is not capable.

More functionality means more complicated use and more complicated implementation. Less functionality is a good thing (if that's enough functionality to solve the problem).

samdark: I don't get your last example :(

kamarton: > You have to do a second read if you want fresh information. There's no way you can avoid that.

simple example caching last result of increment.

> More functionality means more complicated use and more complicated implementation. Less functionality is a good thing (if that's enough functionality to solve the problem).

I don't think it would complicate the process, since the result in `incrementAndGetResult` should be able to be generated anyway.
However, `getResult` itself is also suitable for processes such as having the administrator list the problem rate limiter in the admin UI, eg. supports.

samdark: > simple example: store inside of object the last result of increment.

But then you're not solving anything because information will be equal to the one you get at the time of increment.

kamarton: > But then you're not solving anything because information will be equal to the one you get at the time of increment.

I'm just saying that +1 reading is avoided, and even middleware can be implemented without using it, but depriving the interface is not good.

samdark: @yiisoft/reviewers need more opinions on this issue.

viktorprogger: > I don't get your last example :(

As I see @kamarton suggests to clone counter inside limiter. Am I right?
In my opinion it is a bad decision: user may want to get the exact instance of counter he gave to a limiter and to get its values, serialize it or do anything else. So I think it's better to store interval and limit in counter constructor.

---------------
A bit about `getStatistic`.
In my opinion percentage of cases when this approach is reasonable is too low. @kamarton, can you name some cases when it is critical to handle the situation with different results during one request?

samdark: For now I've decided to merge pull request as is. @kamarton if you want to discuss it further, please create new issue or a forum topic. Thanks.

samdark:
Thank you very much, @romkatsu for implementing it and everyone else for reviews.

Issue #177, opened 2019-11-27
Ability to specify custom templates for error handler HtmlRenderer


samdark:
There should be ability to specify custom template paths for HtmlRenderer.

- `exception` is for verbose mode.
- `error` is for production mode.

strychu:
I've added additional parameters, but maybe it would be more elegant solution if custom path for renderers would be read from config file?

xepozz:
Of course. Can you do it?

strychu:
Ok, i'll work on that

yiiliveext:
We already have the ability to add a custom error handler in the controller action.

xepozz:
This enhancement will be applied to all controllers (I think so)

samdark:
@yiiliveext it's about being able to provide your own templates such as https://github.com/yiisoft/yii-web/blob/master/src/ErrorHandler/HtmlRenderer.php#L38 without changing code of HtmlRenderer  itself.

PR #217, opened 2020-02-11
Custom templates for HtmlRenderer

hstrychalski:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| #177

### DIFF
--- a/src/ErrorHandler/HtmlRenderer.php
+++ b/src/ErrorHandler/HtmlRenderer.php
@@ -7,10 +7,26 @@ use Yiisoft\Yii\Web\Info;
 
 final class HtmlRenderer extends ThrowableRenderer
 {
+    private const DEFAULT_TEMPLATES = [
+        'callStackItem',
+        'error',
+        'exception',
+        'previousException'
+    ];
+
     private int $maxSourceLines = 19;
     private int $maxTraceLines = 13;
 
     private string $traceLine = '{html}';
+
+    private string $errorTemplate;
+    private string $exceptionTemplate;
+
+    public function __construct(array $templates = [])
+    {
+        $this->errorTemplate = isset($templates['error']) ? $templates['error'] : 'error';
+        $this->exceptionTemplate = isset($templates['exception']) ? $templates['exception'] : 'exception';
+    }
 
     public function withMaxSourceLines(int $maxSourceLines): self
     {
@@ -35,14 +51,14 @@ final class HtmlRenderer extends ThrowableRenderer
 
     public function render(\Throwable $t): string
     {
-        return $this->renderTemplate('error', [
+        return $this->renderTemplate($this->errorTemplate, [
             'throwable' => $t,
         ]);
     }
 
     public function renderVerbose(\Throwable $t): string
     {
-        return $this->renderTemplate('exception', [
+        return $this->renderTemplate($this->exceptionTemplate, [
             'throwable' => $t,
         ]);
     }
@@ -54,9 +70,14 @@ final class HtmlRenderer extends ThrowableRenderer
 
     private function renderTemplate(string $template, array $params): string
     {
-        $path = __DIR__ . '/templates/' . $template . '.php';
+        if (in_array($template, self::DEFAULT_TEMPLATES)) {
+            $path = __DIR__ . '/templates/' . $template . '.php';
+        } else {
+            $path = $template;
+        }
+
         if (!file_exists($path)) {
-            throw new \RuntimeException("$template not found at $path");
+            throw new \RuntimeException("$template not found");
         }
 
         $renderer = function (): void {
--- a/src/ErrorHandler/HtmlRenderer.php
+++ b/src/ErrorHandler/HtmlRenderer.php
@@ -12,6 +12,8 @@ final class HtmlRenderer extends ThrowableRenderer
 
     private string $traceLine = '{html}';
 
+    private string $defaultTemplatePath;
+
     private string $errorTemplate;
     private string $exceptionTemplate;
 
@@ -19,9 +21,10 @@ final class HtmlRenderer extends ThrowableRenderer
 
     public function __construct(array $templates = [])
     {
+        $this->defaultTemplatePath = $templates['path'];
         $this->defaultTemplates = $templates['default'];
-        $this->errorTemplate = $templates['error'] ?? 'error';
-        $this->exceptionTemplate = $templates['exception'] ?? 'exception';
+        $this->errorTemplate = $templates['error'] ?? $this->defaultTemplatePath . '/error.php';
+        $this->exceptionTemplate = $templates['exception'] ?? $this->defaultTemplatePath . '/exception.php';;
     }
 
     public function withMaxSourceLines(int $maxSourceLines): self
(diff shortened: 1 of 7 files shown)


src/ErrorHandler/ThrowableRendererInterface.php
```diff
@@ -13,17 +13,21 @@ interface ThrowableRendererInterface
      * Convert throwable into its string representation suitable for displaying in production environment
      *
      * @param \Throwable $t
+     * @param string $template
+     * @param string|null $customPath
      * @return string
      */
-    public function render(\Throwable $t): string;
+    public function render(\Throwable $t, string $template = 'error', string $customPath = null): string;
```
samdark: That would break all other renderers: JsonRenderer, PlainTextRenderer, XmlRenderer. Also, there's no need for template and path for all these. It is HTML-specific.

samdark (changes requested):

src/ErrorHandler/ThrowableRendererInterface.php
strychu: I've moved providing custom path to config file to not change interface methods

tests/ErrorHandler/HtmlRendererTest.php
```diff
@@ -34,4 +40,75 @@ public function testVerboseOutput(): void
...
+        $exception = new \RuntimeException($exceptionMessage);
+
+        $renderedOutput = $renderer->renderVerbose($exception);
+        $this->assertStringContainsString($exceptionMessage, $renderedOutput);
+        $this->assertStringContainsString('<html>', $renderedOutput);
+    }
+
+    public function testRenderTemplateThrowsExceptionWhenTemplateFileNotExists(): void
+    {
+        $exampleNonExistingFile = sprintf('%s.php', bin2hex(random_bytes(16)));
```
xepozz, line 80:
```suggestion
        $exampleNonExistingFile = '_not_found_.php';
```

tests/ErrorHandler/HtmlRendererTest.php
```diff
@@ -34,4 +40,75 @@ public function testVerboseOutput(): void
...
+        $templates = [
+            'error' => $exampleNonExistingFile
+        ];
+
+        $renderer = new HtmlRenderer($templates);
+        $exceptionMessage = 'exception-test-message';
```
xepozz, line 86: Should be removed

tests/ErrorHandler/HtmlRendererTest.php
```diff
@@ -34,4 +40,75 @@ public function testVerboseOutput(): void
...
+        $this->assertStringContainsString('<html>', $renderedOutput);
+    }
+
+    public function testVerboseOutputWithCustomTemplate(): void
+    {
+        $templateFileContents = '<html><?php echo $throwable->getMessage();?></html>';
+        $this->createTestTemplate(self::CUSTOM_TEMPLATES['exception'], $templateFileContents);
+
+        $renderer = new HtmlRenderer(self::CUSTOM_TEMPLATES);
+        $request = new ServerRequest('GET', '/', ['Accept' => ['text/html']]);
```
xepozz, line 67: Please use mock instead of making new instance of external dependency

tests/ErrorHandler/HtmlRendererTest.php
```diff
@@ -34,4 +40,75 @@ public function testVerboseOutput(): void
...
+        $renderer->setRequest($request);
+
+        $exceptionMessage = 'exception-test-message';
+        $exception = new \RuntimeException($exceptionMessage);
+
+        $renderedOutput = $renderer->render($exception);
+        $this->assertStringContainsString($exceptionMessage, $renderedOutput);
+        $this->assertStringContainsString('<html>', $renderedOutput);
```
xepozz, line 58: You have full content of template above. 
Please make sure that `$renderOutput` constain full templates artifacts

tests/ErrorHandler/HtmlRendererTest.php
```diff
@@ -34,4 +40,75 @@ public function testVerboseOutput(): void
...
         $this->assertStringContainsString($exceptionMessage, $renderedOutput);
     }
+
+    public function testNonVerboseOutputWithCustomTemplate(): void
+    {
+        $templateFileContents = '<html><?php echo $throwable->getMessage();?></html>';
+        $this->createTestTemplate(self::CUSTOM_TEMPLATES['error'], $templateFileContents);
+
+        $renderer = new HtmlRenderer(self::CUSTOM_TEMPLATES);
+        $request = new ServerRequest('GET', '/', ['Accept' => ['text/html']]);
```
xepozz, line 50: The same below

xepozz (changes requested):

src/ErrorHandler/HtmlRenderer.php
```diff
@@ -7,11 +7,27 @@
 
 final class HtmlRenderer extends ThrowableRenderer
 {
+    private const DEFAULT_TEMPLATES = [
```
xepozz, line 10: You have `$params['HtmlRenderer']['templates']`, please use it instead

src/ErrorHandler/HtmlRenderer.php
```diff
@@ -54,9 +70,14 @@ private function htmlEncode(string $text): string
...
-        $path = __DIR__ . '/templates/' . $template . '.php';
+        if (in_array($template, self::DEFAULT_TEMPLATES)) {
+            $path = __DIR__ . '/templates/' . $template . '.php';
+        } else {
+            $path = $template;
+        }
+
         if (!file_exists($path)) {
-            throw new \RuntimeException("$template not found at $path");
+            throw new \RuntimeException("Template not found at $path");
```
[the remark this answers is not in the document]
strychu: When providing custom template the $template parameter is actually a path to the template - in this case this message would not be relevant 🤔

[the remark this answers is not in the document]
xepozz: Yes, you are right.

src/ErrorHandler/HtmlRenderer.php
```diff
@@ -54,9 +66,14 @@ private function htmlEncode(string $text): string
 
     private function renderTemplate(string $template, array $params): string
     {
-        $path = __DIR__ . '/templates/' . $template . '.php';
+        if (in_array($template, $this->defaultTemplates)) {
+            $path = __DIR__ . '/templates/' . $template . '.php';
```
xepozz, line 70: Can this path finding be moved into params directly?
This would allow us to get rid of the "default templates"

strychu: Idk if i understand correctly - did you mean to move building default directory path in params? Because inside HtmlRenderer there are functions (i.e. renderCallstackItem) that call renderTemplate only passing template filename so we have to somehow create full path inside HtmlRenderer. 

I can move building default path to params.php so the above line would look sth like:
`$path = $this->defaultTemplatePath . '/' . $template . '.php';` where $this->defaultTemplatePath would be set with with value passed from params in construct.

strychu: Good idea, i didn't like the fact that sometimes there was template name passed and sometimes template path to renderTemplate

samdark (approved):

samdark:
Merged. Thank you!

PR #219, opened 2020-02-18
Add stack to MiddlewareDispatcher

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

You have to change `MiddlewareDispatcherFactory` in `yii-demo` after this PR.
 
```php
class MiddlewareDispatcherFactory
{
    public function __invoke(ContainerInterface $container)
    {
        $session = $container->get(SessionMiddleware::class);
        $router = $container->get(Router::class);
        $errorCatcher = $container->get(ErrorCatcher::class);
        $subFolder = $container->get(SubFolder::class);
        $dispatcher = new MiddlewareDispatcher($container);

        return $dispatcher
            ->addMiddleware($router)
            ->addMiddleware($subFolder)
            ->addMiddleware($session)
            ->addMiddleware($errorCatcher);
    }
}
```

### DIFF
--- a/src/MiddlewareDispatcher.php
+++ b/src/MiddlewareDispatcher.php
@@ -13,7 +13,7 @@ use Yiisoft\Yii\Web\Middleware\Callback;
 /**
  * MiddlewareDispatcher
  */
-final class MiddlewareDispatcher implements RequestHandlerInterface, MiddlewareInterface
+final class MiddlewareDispatcher
 {
     /**
      * @var MiddlewareInterface[]
@@ -30,6 +30,13 @@ final class MiddlewareDispatcher implements RequestHandlerInterface, MiddlewareI
      */
     private $container;
 
+    /**
+     * Contains a chain of middleware wrapped in handlers.
+     * Each handler points to the handler of middleware that will be processed next.
+     * @var RequestHandlerInterface|null stack of middleware
+     */
+    private ?RequestHandlerInterface $stack = null;
+
     public function __construct(
         array $middlewares,
         ContainerInterface $container,
@@ -41,8 +48,8 @@ final class MiddlewareDispatcher implements RequestHandlerInterface, MiddlewareI
 
         $this->container = $container;
 
-        foreach ($middlewares as $middleware) {
-            $this->add($middleware);
+        for ($i = count($middlewares) - 1; $i >= 0; $i--) {
+            $this->addMiddleware($middlewares[$i]);
         }
 
         $responseFactory = $container->get(ResponseFactoryInterface::class);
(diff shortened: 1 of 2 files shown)


samdark (approved):

yiiliveext:
I think we could make the process() method protected.

src/MiddlewareDispatcher.php
```diff
@@ -41,60 +41,70 @@ public function __construct(
 
         $this->container = $container;
 
-        foreach ($middlewares as $middleware) {
-            $this->add($middleware);
+        for ($i = count($middlewares) - 1; $i >= 0; $i--) {
```
viktorprogger: Why did you change `foreach` to `for`?

viktorprogger: Why did you change `foreach` to `for`?

yiiliveext: Hmm... I think this is faster then foreach+array_reverse().

roxblnfk:
I suggest:

`MiddlewareDispatcher` rename to `Pipeline`

For the `process()` method add `MiddlewareInterface`
The `process()` method should throw an exception if the stack of middlewares is empty

The `dispatch()` method rename to `handle()` and add `RequestHandlerInterface`
The `handle()` method should throw an exception if the $nextHandler is not defined

The `addMiddleware()` rename to `add()`
Remove `$middlewares` argument from the `__construct()` method

src/MiddlewareDispatcher.php
```diff
@@ -13,88 +13,89 @@
...
-            $this->middlewares[] = $middleware;
-        } else {
+            $middleware = new Callback($middleware, $this->container);
+        }
+
+        if (!$middleware instanceof MiddlewareInterface) {
             throw new \InvalidArgumentException('Middleware should be either callable or MiddlewareInterface instance. ' . get_class($middleware) . ' given.');
         }
+
+        array_unshift($this->middlewares, $middleware);
```
roxblnfk:
```suggestion
        $this->middlewares[] = $middleware;
```

src/MiddlewareDispatcher.php
```diff
@@ -13,88 +13,89 @@
...
-    public function handle(ServerRequestInterface $request): ResponseInterface
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        $middleware = current($this->middlewares);
-        next($this->middlewares);
-        if ($middleware === false) {
-            if ($this->nextHandler !== null) {
-                return $this->nextHandler->handle($request);
+        if ($this->stack === null) {
+            for ($i = count($this->middlewares) - 1; $i >= 0; $i--) {
```
roxblnfk:
```suggestion
            foreach ($this->middlewares as $middleware) {
```

src/MiddlewareDispatcher.php
```diff
@@ -13,88 +13,89 @@
...
+                $handler = $this->wrap($this->middlewares[$i], $handler);
```
roxblnfk:
```suggestion
                $handler = $this->wrap($middleware, $handler);
```

roxblnfk (changes requested):

src/MiddlewareDispatcher.php
viktorprogger: 1. You don't need `array_reverse` here
1. [Foreach is faster a bit](https://github.com/viktorprogger/foreach-bench) (without `array_reverse`) and more readable: 
image

viktorprogger (changes requested):

src/MiddlewareDispatcher.php
yiiliveext: I thought exactly the same but I so lazy...

src/MiddlewareDispatcher.php
yiiliveext: I thought exactly the same but I so lazy...

src/MiddlewareDispatcher.php
yiiliveext: Done.

samdark:
@roxblnfk nice ideas above but let's handle these separately, OK?

samdark (approved):

samdark:
Thanks!

PR #222, opened 2020-02-20
Added events

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Added events that says that application was started/ended (it needs for debugger)
Fixed bug: `psr/event-dispatcher` is used, but not required in composer


samdark (approved):

samdark:
👍

PR #224, opened 2020-02-21
Refactoring

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

Inlined classes are extracted to regular classes, created tests for them

### DIFF
--- a/src/Middleware/Callback.php
+++ b/src/Middleware/Callback.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Yiisoft\Yii\Web\Middleware;
+
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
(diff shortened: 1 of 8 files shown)


src/Middleware/Callback.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Yiisoft\Yii\Web\Middleware;
+
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Injector\Injector;
+
+final class Callback implements MiddlewareInterface
```
samdark: We've just killed it.  No need to resurrect it.

samdark (changes requested):
Parts that were split aren't usable separately and exposing unnecessary detail to end user.

PR #230, opened 2020-03-17
Add unique request id

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/Middleware/TagRequest.php
+++ b/src/Middleware/TagRequest.php
@@ -19,6 +19,6 @@ final class TagRequest implements MiddlewareInterface
 
     private function getRequestTag(): string
     {
-        return dechex(microtime(true) * 1000000);
+        return $request = uniqid('yii-request-', true);
     }
 }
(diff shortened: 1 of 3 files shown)


samdark:
There's https://github.com/yiisoft/yii-web/blob/master/src/Middleware/TagRequest.php for the purpose.

yiiliveext:
> There's https://github.com/yiisoft/yii-web/blob/master/src/Middleware/TagRequest.php for the purpose.

It is optional. This PR is for another PR.

src/Middleware/TagRequest.php
```diff
@@ -19,6 +19,6 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
     private function getRequestTag(): string
     {
-        return dechex(microtime(true) * 1000000);
+        return $request = uniqid('yii-request-', true);
```
samdark:
```suggestion
        return uniqid('', true);
```

yiiliveext: Done

Issue #218, opened 2020-02-18
Implement AutoLoginMiddleware


samdark:
https://github.com/yiisoft/yii-web/blob/master/src/User/AutoLoginMiddleware.php

mapeveri:
@samdark I would like work in this issue. I have some questions:

1. Right now it already implement the 'remember me' cookie?
2. As I understand if the cookie is there, I get the user and automatically login. Right?
3. Any documentation on how to login?.

Thanks!

samdark:
1. No. There's `AutoLoginMiddleware` but it's empty.
2. Yes. See how it's done in Yii 2 about extra security measures:
  - https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L562
  - https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L584
  - etc.

mapeveri:
@samdark 

1. OK
2. Great
3. I was referring mostly to how yii 3 login works

PR #235, opened 2020-04-01
AutoLoginMiddleware #218

mapeveri:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    |❌
| Tests pass?   | ✔️

This PR is incomplete, I have some questions:

1. I understand that the 'remember' cookie is not defined. Where do i have to define it?
2. For exceptions, what kind of exception do I create?
3. In getIdentityAndDurationFromCookie at https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L562 is there a validateAuthKey method. In yii 3 does this method exist?. How can I trigger warnings?.
4. Is this method in yii 3 exist?.

I think I do not forget anything else, I wait for suggestions and improvements :)


src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+     * Determines if an identity cookie has a valid format and contains a valid auth key.
+     * This method is used when [[enableAutoLogin]] is true.
+     * This method attempts to authenticate a user using the information in the identity cookie.
+     * @param ServerRequestInterface $request Request to handle
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
+     */
+    protected function getIdentityAndDurationFromCookie(ServerRequestInterface $request)
+    {
+        $cookies = $request->getCookieParams();
+        $value = $cookies['remember'] ?? null;
```
roxblnfk, line 56: cookies key shoul be configurable

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+    }
+
+    /**
+     * Determines if an identity cookie has a valid format and contains a valid auth key.
+     * This method is used when [[enableAutoLogin]] is true.
+     * This method attempts to authenticate a user using the information in the identity cookie.
+     * @param ServerRequestInterface $request Request to handle
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
+     */
+    protected function getIdentityAndDurationFromCookie(ServerRequestInterface $request)
```
roxblnfk, line 53:
```suggestion
    protected function getIdentityAndDurationFromCookie(ServerRequestInterface $request): ?array
```

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+    {
+        $cookies = $request->getCookieParams();
+        $value = $cookies['remember'] ?? null;
+
+        if ($value === null) {
+            return null;
+        }
+
+        $data = json_decode($value, true);
+        if (is_array($data) && count($data) == 3) {
```
roxblnfk, line 63:
```suggestion
        if (is_array($data) && count($data) === 3) {
```

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+            list($id, $authKey, $duration) = $data;
```
roxblnfk, line 64:
```suggestion
            [$id, $authKey, $duration] = $data;
```

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+            $identity = $this->identityRepository->findIdentity($id);
```
roxblnfk, line 65: Values from cookies shoul be validate as user data

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $data = $this->getIdentityAndDurationFromCookie($request);
+        if ($this->user->login($data['identity'], $data['duration'])) {
+            try {
+                $response = $handler->handle($request);
+            } catch (\Throwable $e) {
```
xepozz, line 36: Should be removed

src/User/AutoLoginMiddleware.php
roxblnfk: Why `protected`? May be `private`?

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Yiisoft\Auth\IdentityInterface;
+use Yiisoft\Auth\IdentityRepositoryInterface;
+use Yiisoft\Yii\Web\User\User;
+
 /**
  * AutoLoginMiddleware automatically logs user in based on "remember me" cookie
  */
-class AutoLoginMiddleware
+class AutoLoginMiddleware implements MiddlewareInterface
```
roxblnfk, line 17: Final?

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+        $data = $this->getIdentityAndDurationFromCookie($request);
+        if ($this->user->login($data['identity'], $data['duration'])) {
+            try {
+                $response = $handler->handle($request);
+            } catch (\Throwable $e) {
+                throw $e;
+            }
+
+            return $response;
+        } else {
```
xepozz, line 41: Please flip the condition

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+        IdentityRepositoryInterface $identityRepository
+    ) {
         $this->user = $user;
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $data = $this->getIdentityAndDurationFromCookie($request);
+        if ($this->user->login($data['identity'], $data['duration'])) {
```
xepozz, line 33: `$data` may contain `null` value

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+            throw new \Exception("Error authentication");
+        }
+    }
+
+    /**
+     * Determines if an identity cookie has a valid format and contains a valid auth key.
+     * This method is used when [[enableAutoLogin]] is true.
+     * This method attempts to authenticate a user using the information in the identity cookie.
+     * @param ServerRequestInterface $request Request to handle
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
```
xepozz, line 51: It is bad solution to have nullable array return type

src/User/AutoLoginMiddleware.php
roxblnfk: I not see where `$authKey`  is  used

roxblnfk (changes requested):

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
```
xepozz, line 2: `declare(strict_types=1);` is required. Please check other places

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    private $identityInterfaceMock;
+    /**
+     * @var User
+     */
+    private $userMock;
+
+    public function setUp(): void
+    {
+        parent::setUp();
```
xepozz, line 50: Please don't use class properties to mock some objects. 
Mock it directly in method that it needed

src/User/AutoLoginMiddleware.php
xepozz, line 63: Check that `$data` if `false` and do `return` immidiently

src/User/AutoLoginMiddleware.php
```diff
@@ -2,15 +2,76 @@
...
+            return null;
+        }
+
+        $data = json_decode($value, true);
+        if (is_array($data) && count($data) == 3) {
+            list($id, $authKey, $duration) = $data;
+            $identity = $this->identityRepository->findIdentity($id);
+            if ($identity !== null) {
+                if (!$identity instanceof IdentityInterface) {
+                    throw new \Exception("findIdentity() must return an object implementing IdentityInterface.");
```
xepozz, line 68: https://github.com/yiisoft/auth/blob/master/src/IdentityRepositoryInterface.php#L7
It can't be not `IdentityInterface`

xepozz (changes requested):

src/User/AutoLoginMiddleware.php
mapeveri: How would this be? Where do I add the settings?

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: This was in yii-2, I understand that in yii-3 this does not work. It is right?

src/User/AutoLoginMiddleware.php
mapeveri: What method can I use to validate, is there one?

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: What should i remove?

src/User/AutoLoginMiddleware.php
mapeveri: I updated the PR. How is it now, it's okay?

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: What do you propose instead?

tests/User/AutoLoginMiddlewareTest.php
mapeveri: Updated in the last commit

tests/User/AutoLoginMiddlewareTest.php
mapeveri: That is, in each method mock what I need. It doesn't matter if I duplicate it in each method. What do I gain from this?

src/User/AutoLoginMiddleware.php
mapeveri: Return $data ?. No return null?. Or you mean check if it is false and return immediately null.

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

mapeveri:
@xepozz @roxblnfk Thank you very much for the feedback. I pay attention to the comments :)

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,77 @@
...
-    {
+    public function __construct(
+        User $user,
+        IdentityRepositoryInterface $identityRepository
+    ) {
         $this->user = $user;
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
romkatsu, line 32: I think it will be more readable

```php
    {
        if (!$this->userIsAuth($request)) {
            throw new \Exception('Error authentication');
        }

        try {
        } catch (\Throwable $e) {
            throw $e;
        }

        return $response;
    }

    private function userIsAuth(ServerRequestInterface $request): bool
    {
        $data = $this->getIdentityAndDurationFromCookie($request);

        if (is_null($data)) {
            return false;
        }

        if (!$this->user->login($data['identity'], $data['duration'])) {
            return false;
        }

        return true;
    }
```

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,77 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Yiisoft\Yii\Web\User;
 
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
```
romkatsu, line 8: not used

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,98 @@
...
+use PHPUnit_Framework_MockObject_MockObject;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Auth\IdentityInterface;
+use Yiisoft\Auth\IdentityRepositoryInterface;
+use Yiisoft\Yii\Web\User\User;
+use Yiisoft\Yii\Web\User\AutoLoginMiddleware;
+
+class AutoLoginMiddlewareTest extends TestCase
```
romkatsu, line 17: there are places that are not covered by tests
изображение

romkatsu (changes requested):

src/User/AutoLoginMiddleware.php
xepozz: This catching looks useless

src/User/AutoLoginMiddleware.php
xepozz: Use CQRS principle

tests/User/AutoLoginMiddlewareTest.php
xepozz: You can do any useful acts in private methods and combain them like you want

xepozz (changes requested):

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: I know what CQRS is, but I never implemented it. Could you show me a little example? I am interested in improving this. Thank you

tests/User/AutoLoginMiddlewareTest.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Great idea :)

Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

tests/User/AutoLoginMiddlewareTest.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,77 @@
...
+        IdentityRepositoryInterface $identityRepository
+    ) {
         $this->user = $user;
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $data = $this->getIdentityAndDurationFromCookie($request);
+        if (!is_null($data)) {
```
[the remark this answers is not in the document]
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,88 @@
...
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if (!$this->userIsAuth($request)) {
+            throw new \Exception('Error authentication');
+        }
+
+        $response = $handler->handle($request);
```
romkatsu, line 36: Unnecessary local variable 
```php
return $handler->handle($request);
```

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,88 @@
...
+declare(strict_types=1);
+
 namespace Yiisoft\Yii\Web\User;
 
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Yiisoft\Auth\IdentityRepositoryInterface;
+use Yiisoft\Yii\Web\User\User;
```
romkatsu, line 12: not used

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,129 @@
+<?php
+
+
+namespace Yiisoft\Yii\Web\Tests\User;
+
+use Nyholm\Psr7\Response;
+use PHPUnit\Framework\TestCase;
+use PHPUnit_Framework_MockObject_MockObject;
+use Psr\Http\Message\ResponseInterface;
```
romkatsu, line 9: not used

romkatsu (changes requested):

src/User/AutoLoginMiddleware.php
xepozz: It might looks like the following
```php
{
    if ($this->areCredentialsExists($request)) {
        $credentials = $this->parseCredentials($request);

        if (!$this->user->login($credentials['identity'], $credentials['duration'])) {
            throw new \Exception('Auth error');
        }
    }

    return $handler->handle($request);
}

private function areCredentialsExists(ServerRequestInterface $request): bool
{
    return $this->parseCredentials($request) !== [];
}

private function parseCredentials(ServerRequestInterface $request): array
{
    try {
        $data = json_decode($request->getCookieParams(), true, 512, JSON_THROW_ON_ERROR);
    } catch (\JsonException $e) {
        return [];
    }

    if (!is_array($data) || count($data) === 3) {
        return [];
    }

    [$id, , $duration] = $data;
    $identity = $this->identityRepository->findIdentity($id);
    if ($identity === null) {
        return [];
    }

    return ['identity' => $identity, 'duration' => $duration];
}
```

xepozz:
I think we need abstract the credentials storage.
For example, we can store the crentials into 1. cookies, 2. query string, 3. headers, 4. request body, etc.
Later this middleware can be used into `yii-rest` :)

src/User/AutoLoginMiddleware.php
mapeveri: Great, I like this is more readable, I made a mix between this and another proposal

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Some were not used, others if and if I removed them, the tests failed.

tests/User/AutoLoginMiddlewareTest.php
mapeveri: Some were not used, others if and if I removed them, the tests failed.

mapeveri:
I updated my code, although I have a doubt with authKey. Whether it is necessary or not.

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,81 @@
...
+    /**
+     * Parse and determines if an identity cookie has a valid format.
+     * @param ServerRequestInterface $request Request to handle
+     * @return array Returns an array of 'identity' and 'duration' if valid, otherwise [].
+     */
+    private function parseCredentials(ServerRequestInterface $request): array
+    {
+        try {
+            $cookies = $request->getCookieParams();
+            $data = json_decode($cookies['remember'], true, 512);
```
samdark, line 48: Is there `JSON_THROW_ON_ERROR` missing?

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,81 @@
...
+        $identity = $this->identityRepository->findIdentity($id);
+        if ($identity === null) {
+            return [];
+        }
+
+        return ['identity' => $identity, 'duration' => $duration];
+    }
+
+    /**
+     * Check if the user can be authenticated
```
samdark, line 67: That is not what it is doing. It is authenticating user, not just checking.

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,81 @@
...
+     * @param ServerRequestInterface $request Request to handle
+     * @return bool
+     */
+    private function userIsAuth(ServerRequestInterface $request): bool
```
samdark, line 71:
```suggestion
    private function authenticateUserFromRequest(ServerRequestInterface $request): bool
```

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,81 @@
...
+        IdentityRepositoryInterface $identityRepository
+    ) {
         $this->user = $user;
+        $this->identityRepository = $identityRepository;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if (!$this->userIsAuth($request)) {
+            throw new \Exception('Error authentication');
```
samdark, line 33: This behavior will prevent a user with invalid cookie from viewing a website at all. I think we need to behave as in Yii 2 in this case i.e. logging an error: https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L602

samdark (changes requested):

samdark:
As for the questions:

1. It is usually set on logging in with "remember me" checked in the form. So there should be a service that may do it.
2. None.
3. Auth key is required to secure the cookie and allow revoking login stataus. Warnings could be logged with PSR logger interface.
4. There's nothing for auth key yet.

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: Updated in the last commit

src/User/AutoLoginMiddleware.php
mapeveri: This I don't Understand. What is the change?. Thanks!

mapeveri:
@samdark 

1. That is apart from this PR? If not, would you explain a little more.
2. Ok.
3 / 4. In this PR, we are going to do something related to this?

src/User/AutoLoginMiddleware.php
samdark: 1. Require `Psr\Log\LoggerInterface` in constructor.
2. Do `$this->logger->warning('Unable to authenticate used by cookie.');` instead of throwing exception.

samdark:
1. Should be part of the pull request.
3, 4. Yes or else the pull request won't make much sense.

src/User/AutoLoginMiddleware.php
mapeveri: Great, updated in the last commit.

mapeveri:
@samdark ok, great. Can you guide me a little please?.

1. Where I will add the cookie. In the user login method?
2. Related to authkey, I check yii2 to see how it does it?. Any suggestion?

Thanks!

samdark:
1. Yes.
2. Nope. Check Yii 2.

mapeveri:
@samdark I updated my  PR, but I have some questions.

1) I added the getAuthKey, validateAuthKey and sendIdentityCookie in User class, but I don't know how save the cookie. I think I need return the response, but not is clear for me. I need a littler help :)

2) The logout method, I think that need remove the cookie, right?. My doubt is, how to delete the cookie?.

The rest is OK?

Thanks!

samdark:
1. See https://github.com/yiisoft/yii-web/blob/master/src/Cookie.php
2. Yes, we need to delete cookie. Add one with expiration equals `-1`.

src/User/User.php
```diff
@@ -111,6 +113,49 @@ public function setIdentity(IdentityInterface $identity): void
...
+            [
+                $identity->getId(),
+                $this->getAuthKey(),
+                $duration,
+            ],
+            JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE
+        );
+
+        $cookieIdentity = (new Cookie('remember', $data))->expireAt(time() + $duration);
+        $response = new Response();
```
roxblnfk, line 155: If you need to create new Response instance then you should use ResponseFactory from PSR

mapeveri: Can you shared me an example please?. Because I can't find the namespace. Thanks!

mapeveri:
@samdark 

1. What is the equivalent of yii-2 of this method in yii-3: Yii::$app->getResponse()->getCookies()->add($cookie); ??? I am not finding it.

2. Ok, I understand, I updated the PR. 

Thanks!

src/User/User.php
roxblnfk: for example you can explore [#236](https://github.com/yiisoft/yii-web/pull/236/files)

mapeveri: @roxblnfk thanks. I have a question. In the class user I have the sendIdentityCookie method, this is called via method login. In the same class. I have the responseFactory in AutoLoginMiddleware. Is it a good idea to pass the response in the login method as a parameter?

roxblnfk: Wow. I think the User class should not handle either the Response instance or the Request instance.

roxblnfk: In addition, if you do not return a Response object, then after executing the `sendIdentityCookie ()` method, the object will be destroyed

mapeveri: I think the same, that's why I asked the question. I understand if I do not return a Response the cookie won't the save. But I saw yii-2 and I saw this method Yii::$app->getResponse()->getCookies()->add($cookie);  Is there an equivalent?

roxblnfk: You can subscribe to `AfterLogin` and `AfterLogout` events.

mapeveri: @roxblnfk ok, I'm getting closer, but it's still not clear to me. How can I subscribe?. I also don't understand the logic of these events, classes are just getters. Thank you very much.

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,90 @@
...
+        }
+
+        [$id, $authKey, $duration] = $data;
+        $identity = $this->identityRepository->findIdentity($id);
+        if ($identity === null) {
+            return [];
+        }
+
+        if (!$this->user->validateAuthKey($authKey)) {
+            $this->logger->warning('Unable to authenticate used by cookie. Invalid auth key.');
```
xepozz, line 68: But `$this->user->validateAuthKey` just check key, not authenticate user.

src/User/User.php
```diff
@@ -179,6 +228,12 @@ public function logout($destroySession = true): bool
             if ($destroySession && $this->session) {
                 $this->session->destroy();
             }
+
+            // Remove the cookie
+            $expireDateTime = new \DateTime();
```
xepozz, line 233: Please use `DateTimeImmutable`

tests/User/AutoLoginMiddlewareTest.php
xepozz: Not resolved

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,201 @@
...
+        $this->assertEquals($messages[0][1], 'Unable to authenticate used by cookie. Invalid auth key.');
+    }
+
+    public function testProcessCookieEmpty(): void
+    {
+        $this->mockDataRequest();
+        $this->mockDataCookie([]);
+        $this->mockFindIdentity();
+
+        $memory = memory_get_usage();
```
xepozz, line 124: Should be removed. Please check other methods too

tests/User/AutoLoginMiddlewareTest.php
```diff
@@ -0,0 +1,201 @@
...
+    {
+        $this->requestMock
+            ->expects($this->any())
+            ->method('getCookieParams')
+            ->willReturn($cookie);
+    }
+
+    private function mockFindIdentity(): void
+    {
+        $this->identityRepositoryInterfaceMock
```
xepozz, line 173: You can do it directly in mockDataRequest.
Also you don't have to keep properties now

xepozz (changes requested):
Tests might look better

src/User/AutoLoginMiddleware.php
mapeveri: If validation fails, authentication fails. I don't see what is wrong, but did I change the message, do you prefer another?

src/User/User.php
mapeveri: No problem, I updated the PR :)

tests/User/AutoLoginMiddlewareTest.php
mapeveri: I removed the mockFindIdentity. Is it ok?

tests/User/AutoLoginMiddlewareTest.php
mapeveri: I updated this in the last commit.

tests/User/AutoLoginMiddlewareTest.php
xepozz: Yes, but why do you need this line?

tests/User/AutoLoginMiddlewareTest.php
xepozz: How about also remove properties?

tests/User/AutoLoginMiddlewareTest.php
mapeveri: Is true, this line:

$memory = memory_get_usage();

Not is necessary. I updated the PR.

tests/User/AutoLoginMiddlewareTest.php
mapeveri: With identityRepositoryInterfaceMock? What happend with this line. I used this property.

mapeveri:
I would like to finish this PR, but I need a little help to save the cookie in the sendIdentityCookie user method. @roxblnfk told me that I can subscribe to AfterLogin and AfterLogout to get the response. But I don't know how to do this.
Thanks!

src/User/User.php
roxblnfk: Sorry for my absence. As it turned out, now in yii3 you can’t subscribe to events and unsubscribe from it in runtime.
This imposes a limitation on the use of the event system in middleware.
You can try another option: check the status of the user before and after receiving a Response from the next middleware. Based on the status, set cookies in the Response object

mapeveri: @samdark ahh ok I understand that in yii3 not is possible subscribe. But this "check the status of the user before and after receiving a Response from the next middleware. Based on the status, set cookies in the Response object" Not is clear for me. Is there a little example for me to understand?. Thanks!

samdark:
@mapeveri I'll get to it, don't worry. Likely soon.

src/User/User.php
```diff
@@ -179,6 +228,12 @@ public function logout($destroySession = true): bool
             if ($destroySession && $this->session) {
                 $this->session->destroy();
             }
+
+            // Remove the cookie
```
samdark, line 232: This could be added to a method in the Cookie class i.e.

```php
class Cookie
{
    // ...
    public function remove()
    {
        $new = clone $this;
        $new->expire = 1;
        return $new;
    }
```

src/User/User.php
```diff
@@ -111,6 +114,51 @@ public function setIdentity(IdentityInterface $identity): void
...
+                $identity->getId(),
+                $this->getAuthKey(),
+                $duration,
+            ],
+            JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE
+        );
+
+        $expireDateTime = new \DateTimeImmutable();
+        $expireDateTime->setTimestamp(time() + $duration);
+        $cookieIdentity = (new Cookie('remember', $data))->expireAt($expireDateTime);
```
samdark, line 157: It is a good idea to make cookie name configurable.

samdark:
Thanks for working on it!

Middleware seems to be fine. User part needs some more work as I've noted in comments.

src/User/User.php
roxblnfk: Simple example:
```php
/** @var \Yiisoft\Yii\Web\User\User $user */
if ($user->isGuest())
{
    // read cookies
}
// continue pipeline
if (!$user->isGuest())
{
    // write cookies to $response
}
return $response;
```

src/User/User.php
```diff
@@ -2,17 +2,20 @@
...
+use Yiisoft\Yii\Web\Cookie;
 use Yiisoft\Yii\Web\Session\SessionInterface;
+use Yiisoft\Yii\Web\User\AuthenticationKeyInterface;
 use Yiisoft\Yii\Web\User\Event\AfterLogin;
 use Yiisoft\Yii\Web\User\Event\AfterLogout;
 use Yiisoft\Yii\Web\User\Event\BeforeLogin;
 use Yiisoft\Yii\Web\User\Event\BeforeLogout;
 
-class User
+class User implements AuthenticationKeyInterface
```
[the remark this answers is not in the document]
mapeveri: I updated my PR with IdentityInterface.

src/User/User.php
```diff
@@ -111,6 +114,51 @@ public function setIdentity(IdentityInterface $identity): void
...
+                $this->getAuthKey(),
+                $duration,
+            ],
+            JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE
+        );
+
+        $expireDateTime = new \DateTimeImmutable();
+        $expireDateTime->setTimestamp(time() + $duration);
+        $cookieIdentity = (new Cookie('remember', $data))->expireAt($expireDateTime);
+        $response = new Response();
```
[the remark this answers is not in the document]
mapeveri: I updated my PR with this. Is it ok?

src/User/User.php
mapeveri: With this I think we have a problem, because the expiredAt parameters is Datetime. Is it ok this?. By the way, I have a question related to this implementation: Why do they clone, what advantages does it have?

src/User/User.php
mapeveri: I updated my PR with a new property for make this configurable.

src/User/User.php
```diff
@@ -205,6 +261,18 @@ public function getId(): ?string
         return $this->getIdentity()->getId();
     }
 
+    /**
+     * Set the name of the cookie identity
+     * @param string $name New name of the cookie
+     * @return this
+     */
+    public function setIdentityCookie(string $name): this
```
romkatsu, line 269: Undefined class 'this' 

self?

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,91 @@
...
+     * Parse and determines if an identity cookie has a valid format.
+     * @param ServerRequestInterface $request Request to handle
+     * @return array Returns an array of 'identity' and 'duration' if valid, otherwise [].
+     */
+    private function parseCredentials(ServerRequestInterface $request): array
+    {
+        try {
+            $cookies = $request->getCookieParams();
+            $data = json_decode($cookies['remember'], true, 512, JSON_THROW_ON_ERROR);
+        } catch (\Exception $e) {
```
romkatsu, line 53: JsonException?

romkatsu (changes requested):

src/User/User.php
mapeveri: ahahaha is true. I updated in my last commit :)
Thanks!!

src/User/AutoLoginMiddleware.php
mapeveri: I use exception to also encompass getCookieParams. It is not necessary?

src/User/User.php
samdark: `clone` gives you immutability that ensures that there will be no side effects such as:

```php
$cookie = new Cookie();

$cookieToAdd = new CookieCollection();
$cookieToAdd($cookie);

// if in saveCookie() we modify the object, it will end up modified in collection
// usually that's not what we want
$anotherClass->modifyAndSaveCookie($cookie);

$andAnotherClass->sendCookies($cookie);
```

samdark: Cookies are going to be adjusted: https://github.com/yiisoft/yii-web/pull/244

Overall the idea with `remove()` is about setting date in the past. In the pull request mentioned there's such a method called `expire()`.

mapeveri: @samdark ahh ok thanks for explaining. 

Regarding the elimination of cookies, we hope to merge that PR, right?. I updated this PR assuming the expired method exists. But I don't sure if is correctly.

src/User/AutoLoginMiddleware.php
romkatsu: Ok

src/User/User.php
samdark: Yes, it will be merged.

mapeveri: @samdark Ok, I understand that this PR would be already, right?

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,91 @@
...
+        LoggerInterface $logger
+    ) {
         $this->user = $user;
+        $this->identityRepository = $identityRepository;
+        $this->logger = $logger;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
```
samdark, line 35: Shouldn't we try to log user in _before_ handling request?

src/User/User.php
```diff
@@ -3,16 +3,18 @@
...
 use Yiisoft\Auth\IdentityRepositoryInterface;
+use Yiisoft\Yii\Web\Cookie;
 use Yiisoft\Yii\Web\Session\SessionInterface;
 use Yiisoft\Yii\Web\User\Event\AfterLogin;
 use Yiisoft\Yii\Web\User\Event\AfterLogout;
 use Yiisoft\Yii\Web\User\Event\BeforeLogin;
 use Yiisoft\Yii\Web\User\Event\BeforeLogout;
 
-class User
+class User implements IdentityInterface
```
samdark, line 17: [That's not correct](https://github.com/yiisoft/yii-web/pull/235#discussion_r409841382). That's not an identity.

samdark (changes requested):

src/User/AutoLoginMiddleware.php
mapeveri: It is true! I have a question, how can I get the response? For example DI in the method process?

src/User/User.php
mapeveri: But here https://github.com/yiisoft/yii-demo/blob/master/src/Entity/User.php#L29 use the IdentityInterface. Not is clear for me. Thanks!

samdark: Yes. The interface implementation should be done in an entity and should _not_ be done in the service in this pull request. The service should use the entity.

mapeveri: Ok perfect i will remove the interface!

src/User/AutoLoginMiddleware.php
samdark: You don't need response in this case but, in general, you create it via response factory. See https://github.com/yiisoft/yii-web/blob/master/src/Middleware/Csrf.php#L40

samdark: Aha, I think I've got what you've meant...

samdark: The method body should be something like:

```php
if (!$this->authenticateUserFromRequest($request, $response)) {
    $this->logger->warning('Unable to authenticate user by cookie.');
}

return $handler->handle($request);
```

mapeveri: But I need add the response for save the cookie. I will add the response with the responseFactory, but i have a question. This new response it is different to the response of the handler, right?. Is it a problem?  

```
$response = $this->responseFactory->createResponse(Status::UNPROCESSABLE_ENTITY);
if (!$this->authenticateUserFromRequest($request, $response)) {
    $this->logger->warning('Unable to authenticate user by cookie.');
}

return $handler->handle($request);
```

samdark: > This new response it is different to the response of the handler, right?. Is it a problem?

Yes, that's a problem.

samdark: It seems like sending identity cookie should be done separately from logging in by cookie.

mapeveri: How would this be?

src/User/User.php
mapeveri: I updated the PR with this!

src/User/User.php
```diff
@@ -128,13 +176,15 @@ public function setIdentity(IdentityInterface $identity): void
      *
      * @param IdentityInterface $identity the user identity (which should already be authenticated)
      * @param int $duration number of seconds that the user can remain in logged-in status, defaults to `0`
+     * @param ResponseInterface $response Response to handle
      * @return bool whether the user is logged in
      */
-    public function login(IdentityInterface $identity, int $duration = 0): bool
+    public function login(IdentityInterface $identity, int $duration = 0, ResponseInterface $response): bool
```
samdark, line 182: Setting response here isn't a good idea overall. I think it is middleware job.

mapeveri: If I remove the response out of there, I have no way to add the cookie. I could add the response as a parameter of the middleware process method, that way I get the response and I can pass it to the login so that it saves the cookie. Now if I remove that parameter from the login method, I can't do that.

samdark:
@mapeveri is it OK for you if I'll get in and change things directly in this pull request?

mapeveri:
@mapeveri no problem my friend. Then I would like to see the changes, to have it in mind for the future :)

samdark:
TODO:

- [ ] Add tests for `AutoLogin`.
- [ ] Decide if `AutoLogin` should react to `AfterLogin` and `AfterLogout`. Implement if the decision is positive.

src/User/AutoLogin.php
```diff
@@ -0,0 +1,64 @@
...
+ * The service is used to send or remove auto-login cookie.
+ *
+ * @see AutoLoginIdentityInterface
+ * @see AutoLoginMiddleware
+ */
+class AutoLogin
+{
+    private string $cookieName = 'autoLogin';
+
+    public function cookieName(string $name): self
```
yiiliveext, line 20: withCookieName() will be better

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,89 @@
...
+        if (!$identity instanceof AutoLoginIdentityInterface) {
+            throw new \RuntimeException('Identity repository must return an instance of \Yiisoft\Yii\Web\User\AutoLoginIdentityInterface in order for auto-login to function.');
+        }
+
+        if (!$identity->validateAuthKey($authKey)) {
+            $this->logger->warning('Unable to authenticate user by cookie. Invalid auth key.');
+            return;
+        }
+
+        if ($identity !== null) {
```
yiiliveext, line 85: This condition is unnecessary

yiiliveext:
> * [ ]  Decide if `AutoLogin` should react to `AfterLogin` and `AfterLogout`. Implement if the decision is positive.
+

src/User/AutoLogin.php
```diff
@@ -0,0 +1,66 @@
+<?php
```
romkatsu, line 1: declare strict types please

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,99 @@
...
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $this->authenticateUserByCookieFromRequest($request);
+        $guestBeforeHandle = $this->user->isGuest();
+        $response = $handler->handle($request);
+        $guestAfterHandle = $this->user->isGuest();
+
+        if ($guestBeforeHandle && !$guestAfterHandle) {
+            $this->autoLogin->addCookie($this->user->getIdentity(false), $response);
```
romkatsu, line 44: not enough tests

изображение

romkatsu (changes requested):

samdark:
TODO:

Create an issue about adding two tests `testAddCookieAfterLogin()` and `testRemoveCookieAfterLogout()`. No idea how to implement these in a good way :(

samdark:
@mapeveri it's mostly finished except two tests that I have no idea on how to implement in a good way. What do you think?

rustamwin (approved):

src/User/Event/AfterLogin.php
```diff
@@ -7,21 +7,14 @@
 class AfterLogin
```
roxblnfk, line 7: declare strict?

src/User/Event/BeforeLogin.php
```diff
@@ -7,13 +7,11 @@
 class BeforeLogin
```
roxblnfk, line 7: declare strict?

src/User/AutoLoginMiddleware.php
```diff
@@ -1,16 +1,99 @@
...
+
+        return $response;
+    }
+
+    /**
+     * Authenticate user by auto login cookie from request.
+     *
+     * @param ServerRequestInterface $request Request instance containing auto login cookie.
+     */
+    private function authenticateUserByCookieFromRequest(ServerRequestInterface $request): void
```
roxblnfk, line 59: What do you think of breaking this large method into several small ones?

roxblnfk (approved):

src/User/AutoLoginMiddleware.php
samdark: I don't have a good way to do so.

samdark:
Thanks for starting it, @mapeveri.

mapeveri:
@samdark thanks to you. I would like make a some questions, for understand, because I have some doubts. The if of the line https://github.com/mapeveri/yii-web/blob/e89d374cf0b9d2dc838ac449f46e9dde510d963a/src/User/AutoLoginMiddleware.php#L43

and the line https://github.com/mapeveri/yii-web/blob/e89d374cf0b9d2dc838ac449f46e9dde510d963a/src/User/AutoLoginMiddleware.php#L47

I don't understands this. It is not clear to me when each condition is met to add the cookie or not. The login is done on this line: https://github.com/mapeveri/yii-web/blob/e89d374cf0b9d2dc838ac449f46e9dde510d963a/src/User/AutoLoginMiddleware.php#L97

samdark:
These conditions are to detect if user logged in (was guest, became non-guest) or logged out (was non-guest, became guest) in the current request. On login we're setting a cookie. On logout we're removing a cookie. This is totally separated from logging user in by existing cookie that it happening regardless in the very beginning.

mapeveri:
@samdark thanks!!

PR #236, opened 2020-04-01
Middleware for redirection to HTTPS 

roxblnfk:
Added `ForceSecureConnection` middleware for redirection to secure connetion

* [x]  Middleware code
* [x]  Configurable
* [ ] ~~Add CSP reporting~~
* [x] Docs
* [x] Tests


src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Http\Status;
+
+/**
+ * Redirects from HTTP to HTTPS and adds CSP and HSTS headers
```
samdark, line 15: A note should be added that it's better to do it with webserver config. See https://github.com/yiisoft/yii-web/issues/237

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+class ForceSecureConnection implements MiddlewareInterface
```
samdark, line 17: final?

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+{
+    protected int $statusCode = Status::MOVED_PERMANENTLY;
```
samdark, line 19: private? Also, I think there should be a `withStatusCode` or alike method.

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+    protected ?int $port = null;
```
samdark, line 20: `withPort` is needed, right?

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+
+    protected bool $addCSP = true;
```
samdark, line 22: Should be configurable.

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+    protected string $cspDirective = 'upgrade-insecure-requests';
```
samdark, line 23: Should be configurable.

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+
+    protected bool $addSTS = true;
```
samdark, line 25: Should be configurable.

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,69 @@
...
+    protected int $stsMaxAge = 31_536_000; // 12 months
+    protected bool $stsSubDomains = false;
```
samdark, line 27: Should be configurable.

samdark (changes requested):

src/Middleware/ForceSecureConnection.php
roxblnfk: At the moment, I have not finally decided how this class will be configured. Therefore, all properties are protected, and the class is not final

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,135 @@
...
+
+    private bool $addCSP = true;
+    private string $cspDirectives = self::DEFAULT_CSP_DIRECTIVES;
+
+    private bool $addSTS = true;
+    private int $hstsMaxAge = self::DEFAULT_HSTS_MAX_AGE;
+    private bool $hstsSubDomains = false;
+
+    private ResponseFactoryInterface $responseFactory;
+    private const DEFAULT_CSP_DIRECTIVES = 'upgrade-insecure-requests; default-src https:';
```
rustamwin, line 34: Constants must be declared first

rustamwin (changes requested):

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,138 @@
...
+    private ResponseFactoryInterface $responseFactory;
+
+    public function __construct(ResponseFactoryInterface $responseFactory)
+    {
+        $this->responseFactory = $responseFactory;
+    }
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if ($this->redirect && strcasecmp($request->getUri()->getScheme(), 'http') === 0) {
+            $url = (string)$request->getUri()->withScheme('https')->withPort($this->port);
```
samdark, line 52: Would this overwrite port regardless if `$this->port` is set? Was it intended?

src/Middleware/ForceSecureConnection.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Add Strict-Transport-Security header to each Response
+     * @see Header::STRICT_TRANSPORT_SECURITY
+     * @param int $maxAge
+     * @param bool $subDomains
+     * @return self
+     */
+    public function withHSTS(int $maxAge = self::DEFAULT_HSTS_MAX_AGE, bool $subDomains = false): self
```
samdark, line 110: How about separating hstsSubDomains into separate method?

roxblnfk: The `includeSubDomains` directive is optional, while` max-age` is mandatory. Both directives are indicated on one line of the header.
At the same time, it makes no sense to configure the `includeSubDomains` directive if HSTS turned off (separately from the `withHSTS()` method)

src/Middleware/ForceSecureConnection.php
roxblnfk: If `$this->port` is `null` then URI will be without port like `https://my-site.org/index.php`. In this case Client will use default port for this protocol (443)

Otherwise the port will be part of the URI. And if we manually set the port value as 443 then URI will look like `https://my-site.org:443/index.php`

samdark (approved):

samdark:
All good. Merged. Thank you!

PR #240, opened 2020-04-07
Fix session

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | yiisoft/yii-demo#77


samdark:
Thanks!

TianNex:
thank you for the fix,
but , i am not sure it fixed this problem, it seems the error still happens,

i tried to adjust the sequence of the middleware 
in   \src\Factory\MiddlewareDispatcherFactory.php
it seems OK now,

return (new MiddlewareDispatcher($container))
                        ->addMiddleware($session)
                        ->addMiddleware($csrf)
                        ->addMiddleware($router)
                        ->addMiddleware($subFolder)
                        ->addMiddleware($errorCatcher);

samdark:
@TianNex how to reproduce it? I've tested it for the case in issue description and it works with the fix.

TianNex:
> @TianNex how to reproduce it? I've tested it for the case in issue description and it works with the fix.

sorry,it's OK now, it maybe caused by cache or the old session.php files of yii2
after i clear the files and the file ,it's OK now ,thank you

Issue #86, opened 2019-07-07
Add cookie protection


samdark:
- [ ] Decide if we still need cookie signing with MAC: https://github.com/yiisoft/security/blob/master/src/Mac.php
- [ ] If yes, add it to https://github.com/yiisoft/yii-web/blob/master/src/Cookie.php.
- [ ] Decide on how to retrieve value w/o signature from request.

armpogart:
Would you accept PR for PSR-15 cookie encryption middleware? I have a baked solution for our project already on yii3.

samdark:
Encryption isn't the same as MAC-signing. What's the use case for encrypting a cookie?

armpogart:
Yep, it's true. But they are two things that are sometimes used exclusively for the same purpose, or even as a best practice used both together. Encryption is used when you have somewhat sensitive data you need to store in cookies (by saying somewhat I mean the data that is still ok to store in cookies as really sensitive data must not be stored in cookies). Also encryption of cookies can be sometimes be used as an alternative of signing, as the data will be malformed if somebody tampers it on client side. But ideally they are used together first the data is encrypted and then the data is signed. Cookie encryption middleware can also have a key exclusion map, to define the keys to be excluded during encryption, so that they are available unencrypted on client side while still having other more sensitive values encrypted.

As they are very similar in purpose and functionally, I will be happy to implement both cookie signing and encryption PSR-15 middlewares and try to document their usage (right order).

samdark:
OK.

PR #244, opened 2020-04-15
Refactor Cookie, add CookieCollection

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

Some preliminary refactoring and additional class (`CookieCoolection`) needed for #86 implementation.
Also I think changes are useful even without #86.

The `Cookie` class property defaults were changed to standard cookie defaults so that minimal valid cookie can be created (containing only name and empty value) and parsed by this class and recommended/useful defaults are now set in constructor with flag switch (enabled by default).

### DIFF
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -1,4 +1,5 @@
 <?php
+
 declare(strict_types=1);
 
 namespace Yiisoft\Yii\Web;
@@ -25,9 +26,18 @@ use function strtolower;
  */
 final class Cookie
 {
-    // @see https://tools.ietf.org/html/rfc6265#section-4
-    // @see https://tools.ietf.org/html/rfc2616#section-2.2
+    /**
+     * Regular Expression used to validate cookie name
+     * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
+     * @see https://tools.ietf.org/html/rfc2616#section-2.2
+     */
     private const TOKEN = '/^[a-zA-Z0-9!#$%&\' * +\- .^_`|~]+$/';
+
+    /**
+     * Regular expression used to validate cooke value
+     * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
+     * @see https://tools.ietf.org/html/rfc2616#section-2.2
+     */
     private const OCTET='/^[\x21\x23-\x2B\x2D-\x3A\x3C-\x5B\x5D-\x7E]*$/';
 
     /**
@@ -58,13 +68,18 @@ final class Cookie
     /**
      * @var string value of the cookie
      */
-    private string $value = '';
+    private string $value;
 
     /**
-     * @var string|null RFC-1123 date at which the cookie expires.
+     * @var DateTimeInterface|null RFC-1123 date at which the cookie expires.
      * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
      */
-    private ?string $expire = null;
+    private ?DateTimeInterface $expire = null;
+
+    /**
+     * @var int|null maximum age of cookie in seconds
+     */
+    private ?int $maxAge = null;
 
     /**
      * @var string|null domain of the cookie.
@@ -121,17 +136,32 @@ final class Cookie
         return $this->value;
     }
 
+    public function isExpired(): bool
+    {
+        return isset($this->expire) && $this->expire->getTimestamp() < (new DateTimeImmutable())->getTimestamp();
+    }
+
     public function expireAt(DateTimeInterface $dateTime): self
     {
         $new = clone $this;
-        $new->expire = $dateTime->format('D, d M Y H:i:s T');
+        $new->expire = $dateTime;
         return $new;
     }
 
-    public function validFor(DateInterval $dateInterval): self
+    public function expire(): self
     {
-        $expireDateTime = (new DateTimeImmutable())->add($dateInterval);
-        return $this->expireAt($expireDateTime);
+        return $this->expireAt((new \DateTimeImmutable('-5 years')));
+    }
+
+    public function maxAge(DateInterval $dateInterval): self
+    {
+        $reference = new DateTimeImmutable();
+        $expireDateTime = $reference->add($dateInterval);
+
+        $new = clone $this;
+        $new->expire = $expireDateTime;
+        $new->maxAge = $expireDateTime->getTimestamp() - $reference->getTimestamp();
+        return $new;
     }
 
     public function expireWhenBrowserIsClosed(): self
@@ -180,12 +210,15 @@ final class Cookie
             throw new InvalidArgumentException('sameSite should be one of "Lax", "Strict" or "None"');
         }
 
+        $new = clone $this;
+
         if ($sameSite === self::SAME_SITE_NONE) {
             // the secure flag is required for cookies that are marked as 'SameSite=None'
-            $this->secure = true;
+            // so that cross-site cookies can only be accessed over HTTPS
+            // without it cookie will not be available for external access
+            $new->secure = true;
         }
 
-        $new = clone $this;
         $new->sameSite = $sameSite;
         return $new;
     }
@@ -212,7 +245,12 @@ final class Cookie
         ];
 
         if ($this->expire) {
-            $cookieParts[] = 'Expires=' . $this->expire;
+            $cookieParts[] = 'Expires=' . $this->expire->format(DateTimeInterface::RFC7231);
+        }
+
+        // max-age must be non-zero digit
+        if ($this->maxAge && $this->maxAge > 0) {
+            $cookieParts[] = 'Max-Age=' . $this->maxAge;
         }
 
         if ($this->domain) {
@@ -274,7 +312,7 @@ final class Cookie
                     $cookie = $cookie->expireAt(new DateTimeImmutable($attributeValue));
                     break;
                 case 'max-age':
-                    $cookie = $cookie->validFor(new DateInterval('PT' . $attributeValue . 'S'));
+                    $cookie = $cookie->maxAge(new DateInterval('PT' . $attributeValue . 'S'));
                     break;
                 case 'domain':
                     $cookie = $cookie->domain($attributeValue);
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -20,37 +20,39 @@ use function in_array;
 use function preg_match;
 use function preg_split;
 use function strtolower;
+use function time;
 
 /**
- * Cookie helps adding Set-Cookie header response in order to set cookie
+ * Represent a cookie and also helps adding Set-Cookie header response in order to set cookie
  */
 final class Cookie
 {
     /**
      * Regular Expression used to validate cookie name
-     * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
-     * @see https://tools.ietf.org/html/rfc2616#section-2.2
+     * @link https://tools.ietf.org/html/rfc6265#section-4.1.1
+     * @link https://tools.ietf.org/html/rfc2616#section-2.2
      */
-    private const TOKEN = '/^[a-zA-Z0-9!#$%&\' * +\- .^_`|~]+$/';
+    private const PATTERN_TOKEN = '/^[a-zA-Z0-9!#$%&\' * +\- .^_`|~]+$/';
 
     /**
      * Regular expression used to validate cooke value
-     * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
-     * @see https://tools.ietf.org/html/rfc2616#section-2.2
+     * @link https://tools.ietf.org/html/rfc6265#section-4.1.1
+     * @link https://tools.ietf.org/html/rfc2616#section-2.2
      */
-    private const OCTET='/^[\x21\x23-\x2B\x2D-\x3A\x3C-\x5B\x5D-\x7E]*$/';
+    private const PATTERN_OCTET='/^[\x21\x23-\x2B\x2D-\x3A\x3C-\x5B\x5D-\x7E]*$/';
 
     /**
-     * SameSite policy Lax will prevent the cookie from being sent by the browser in all cross-site browsing context
+     * SameSite policy `Lax` will prevent the cookie from being sent by the browser in all cross-site browsing contexts
      * during CSRF-prone request methods (e.g. POST, PUT, PATCH etc).
      * E.g. a POST request from https://otherdomain.com to https://yourdomain.com will not include the cookie, however a GET request will.
-     * When a user follows a link from https://otherdomain.com to https://yourdomain.com it will include the cookie
+     * When a user follows a link from https://otherdomain.com to https://yourdomain.com it will include the cookie.
+     * This is the default value in modern browsers.
      * @see $sameSite
      */
     public const SAME_SITE_LAX = 'Lax';
 
     /**
-     * SameSite policy Strict will prevent the cookie from being sent by the browser in all cross-site browsing context
+     * SameSite policy `Strict` will prevent the cookie from being sent by the browser in all cross-site browsing contexts
      * regardless of the request method and even when following a regular link.
      * E.g. a GET request from https://otherdomain.com to https://yourdomain.com or a user following a link from
      * https://otherdomain.com to https://yourdomain.com will not include the cookie.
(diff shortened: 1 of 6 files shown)


src/Cookie.php
```diff
@@ -1,14 +1,35 @@
...
+use function preg_match;
+use function preg_split;
+use function strtolower;
+
 /**
  * Cookie helps adding Set-Cookie header response in order to set cookie
  */
 final class Cookie
 {
+    // @see https://tools.ietf.org/html/rfc6265#section-4
```
rustamwin, line 28: Should be phpdoc

src/Cookie.php
```diff
@@ -37,91 +58,103 @@ final class Cookie
     /**
      * @var string value of the cookie
      */
-    private string $value;
+    private string $value = '';
```
xepozz, line 61: It shouldn't have default value

src/Cookie.php
```diff
@@ -37,91 +58,103 @@ final class Cookie
...
+    {
+        return $this->value;
+    }
+
+    public function expireAt(DateTimeInterface $dateTime): self
     {
-        $expireDateTime = (new \DateTimeImmutable())->add($dateInterval);
         $new = clone $this;
-        $new->expire = (int)$expireDateTime->format('U');
+        $new->expire = $dateTime->format('D, d M Y H:i:s T');
```
xepozz, line 127: Would be nice to extract format template as a private constant

src/Cookie.php
```diff
@@ -37,91 +58,103 @@ final class Cookie
...
         return $new;
     }
 
-    public function expireAt(\DateTimeInterface $dateTime): self
+    public function validFor(DateInterval $dateInterval): self
```
xepozz, line 131: Is it really needed?

src/Cookie.php
```diff
@@ -144,37 +177,131 @@ public function httpOnly(bool $httpOnly): self
     public function sameSite(string $sameSite): self
     {
         if (!in_array($sameSite, [self::SAME_SITE_LAX, self::SAME_SITE_STRICT, self::SAME_SITE_NONE], true)) {
-            throw new \InvalidArgumentException('sameSite should be either Lax or Strict');
+            throw new InvalidArgumentException('sameSite should be one of "Lax", "Strict" or "None"');
+        }
+
+        if ($sameSite === self::SAME_SITE_NONE) {
+            // the secure flag is required for cookies that are marked as 'SameSite=None'
+            $this->secure = true;
```
xepozz, line 185: Shouldn't you set `secure` to `$new` object?

src/Cookie.php
```diff
@@ -37,91 +58,103 @@ final class Cookie
...
     }
 
-    public function expireAt(\DateTimeInterface $dateTime): self
+    public function validFor(DateInterval $dateInterval): self
+    {
+        $expireDateTime = (new DateTimeImmutable())->add($dateInterval);
+        return $this->expireAt($expireDateTime);
+    }
+
+    public function expireWhenBrowserIsClosed(): self
```
xepozz, line 137: Is it really needed? We have already `null` value when make new instance of Cookie

src/Cookie.php
roxblnfk: `\DateTimeInterface::RFC7231`

src/Cookie.php
armpogart: Yep, default value here can be omitted as constuctor sets it to default empty string anyway. Empty string is used instead of previous required value, as cookie can have empty value according to RFC.

src/Cookie.php
armpogart: @roxblnfk I first used that constant, but it produces different result than the one defined in RFC, the result of using that constant is e.g. `Sun, 06 Nov 1994 08:49:37 +0000` but according to [RFC](https://tools.ietf.org/html/rfc2616#section-3.3.1) we need `Sun, 06 Nov 1994 08:49:37 GMT`. I couldn't find any information in RFC whether the first format will be ok, so I've used the custom format.

@xepozz Will extract it to constant with link to rfc specification.

src/Cookie.php
armpogart: It is just a helper method. Not sure we need it or not as we can achieve the same with `expireAt`.

src/Cookie.php
armpogart: Yep, my mistake will push a fix with further commit. The `Cookie` is immutable, so we don't change any property directly on it.

src/Cookie.php
armpogart: I see this one as a way to modify already existing cookie and removing expire header from it. This behavior can't be achieved by setting expire date to anything else. The only other way would be to create a new cookie, set all properties accordingly without setting expiry.

src/Cookie.php
armpogart: @xepozz I will rename the method to `maxAge` and it will set `Max-Age` header and also for compatibility reason and it will set `Expire` header to the same value as some agents/browsers don't support (read) `Max-Age` header, but it is present in specification. The behavior of the method will be the same.

src/Cookie.php
armpogart: @roxblnfk It occurs that `\DateTimeInterface::RFC7231` works correctly and tests just failed when I used it.

roxblnfk: RFC7231 is the preferred date and time format in modern HTTP headers

src/Cookie.php
roxblnfk: And it should be `@link` annotation instead of `@see`

src/Cookie.php
```diff
@@ -40,88 +71,120 @@ final class Cookie
...
-     * ```
-     * @see https://www.owasp.org/index.php/SameSite for more information about sameSite.
+     * @var string|null SameSite prevents the browser from sending this cookie along with cross-site requests.
+     * @see https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#samesite-cookie-attribute for more information about sameSite.
      */
-    private string $sameSite = self::SAME_SITE_LAX;
+    private ?string $sameSite = null;
 
-    public function __construct(string $name, string $value)
+    public function __construct(string $name, string $value = '', bool $safeDefaults = true)
```
samdark, line 112: What is the reason for allowing unsafe defaults? I think we should always start with safe defaults allowing to explicitly change them to unsafe ones.

src/Cookie.php
```diff
@@ -40,88 +71,120 @@ final class Cookie
...
-    private ?string $domain = null;
+    private ?DateTimeInterface $expire = null;
 
     /**
-     * @var int the timestamp at which the cookie expires. This is the server timestamp.
-     * Defaults to 0, meaning "until the browser is closed".
+     * @var int|null maximum age of cookie in seconds
      */
-    private int $expire = 0;
+    private ?int $maxAge = null;
```
samdark, line 82: I don't think we need to store both maxAge and expire. One could be obtained from another on header generation.

As for headers it's better to always send both max age and expires (for HTTP 1.0 clients) that result in the same expiration time.

src/Cookie.php
```diff
@@ -40,88 +71,120 @@ final class Cookie
...
      */
-    private string $path = '/';
+    private ?string $domain = null;
 
     /**
-     * @var bool whether cookie should be sent via secure connection
+     * @var string|null the path on the server in which the cookie will be available on.
      */
-    private bool $secure = true;
+    private ?string $path = null;
```
samdark, line 92: I'd make `/` default since it's what majority users expect and add `withoutPath()` method for choosing not to send it in the header.

src/Cookie.php
```diff
@@ -40,88 +71,120 @@ final class Cookie
...
 
     /**
-     * @var bool whether the cookie should be accessible only through the HTTP protocol.
+     * @var bool|null whether cookie should be sent via secure connection
```
samdark, line 95:
```suggestion
     * @var bool|null whether cookie should be sent via secure connection only
```

src/Cookie.php
```diff
@@ -40,88 +71,120 @@ final class Cookie
...
+    }
+
+    public function getValue(): string
+    {
+        return $this->value;
+    }
+
+    public function isExpired(): bool
+    {
+        return isset($this->expire) && $this->expire->getTimestamp() < (new DateTimeImmutable())->getTimestamp();
```
samdark, line 141:
```suggestion
        return $this->expire !== null && $this->expire->getTimestamp() < (new DateTimeImmutable())->getTimestamp();
```

src/Cookie.php
```diff
@@ -144,37 +207,139 @@ public function httpOnly(bool $httpOnly): self
...
+     * Pass only Set-Cookie header value.
+     *
+     * @param string $string 'Set-Cookie' header value
+     * @return self
+     * @throws Exception
+     */
+    public static function fromSetCookieString(string $string): self
+    {
+        // array_filter with empty callback is used to filter out all falsy values
+        $rawAttributes = array_filter(preg_split('~\s*[;]\s*~', $string));
```
samdark, line 290: Can we use PREG_SPLIT_NO_EMPTY?

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+
+        return new self($elements);
+    }
+
+    /**
+     * Populates the cookie collection from a ResponseInterface.
+     * @param ResponseInterface $response the response object to populate from
+     * @return self collection created from response
+     */
+    public static function fromResponse(ResponseInterface $response): self
```
samdark, line 289: What is the use case for it?

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web;
+
+use ArrayAccess;
+use ArrayIterator;
+use Closure;
+use Countable;
+use DateTimeImmutable;
```
samdark, line 11: Not used.

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+use InvalidArgumentException;
+use IteratorAggregate;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
```
samdark, line 15: Not used.

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+                throw new InvalidArgumentException('CookieCollection can contain only Cookie instances.');
+            }
+
+            $this->cookies[$cookie->getName()] = $cookie;
+        }
+    }
+
+    /**
+     * Returns the collection as a PHP array.
+     * The array keys are cookie names, and the array values are the corresponding cookie objects.
```
samdark, line 49: Could be type-hinted as `@return Cookie[]`.

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+    {
+        return isset($this->cookies[$name]) ? $this->cookies[$name]->getValue() : $defaultValue;
+    }
+
+    /**
+     * Adds a cookie to the collection.
+     * If there is already a cookie with the same name in the collection, it will be removed first.
+     * @param Cookie $cookie the cookie to be added
+     */
+    public function add(Cookie $cookie): void
```
samdark, line 156: What's the goal of adding a cookie to it? Is `addToResponse(ResponseInterface $response): ResponseInterface` planned?

src/CookieCollection.php
```diff
@@ -0,0 +1,298 @@
...
+
+use function array_keys;
+use function array_values;
+use function count;
+use function in_array;
+
+/**
+ * A CookieCollection is a collection implementation to wrap an array of Cookie class instances
+ */
+final class CookieCollection implements IteratorAggregate, ArrayAccess, Countable
```
samdark, line 25: The use case overall isn't clear.

samdark (changes requested):

src/Cookie.php
```diff
@@ -1,14 +1,45 @@
...
+use function strtolower;
+
 /**
  * Cookie helps adding Set-Cookie header response in order to set cookie
  */
 final class Cookie
 {
+    /**
+     * Regular Expression used to validate cookie name
+     * @see https://tools.ietf.org/html/rfc6265#section-4.1.1
```
roxblnfk, line 31: All `@see` annotations with url to RFC should be replaced to `@link`

src/Cookie.php
armpogart: See [Cookie::fromSetCookieString](https://github.com/yiisoft/yii-web/blob/439db1b2fd5befe9a59e2315dcf2ad468f79f7a7/src/Cookie.php#L262), it creates a cookie without any defaults and later populates it from the string. If we don't have such an option, we would need to create a cookie and then chain several methods to reset the defaults while also creating several objects in the process and deleting them (all `clone` statements) and overload GC unnecessarily during that.

Also it is reasonable to allow creating minimal valid cookie without any defaults without any hassle.

src/Cookie.php
armpogart: Yep, I agree. But I see one problem here. A valid cookie can have different expiry and maxAge. And according to RFC the maxAge will have precedence in that case. If we don't store maxAge and generate from expiry.
What about cookies that are not set by yii backend, cookies that we just read with `Cookie::fromSetCookieString` if we don't store maxAge and expiry separately, and cookie has both set on client side, we would just change that cookie with in an unexpected way for the user. What if some frontend or implementation somehow depends on it. I'm sure that such behavior (setting different expiry and maxAge) is wrong at all, but it is valid per RFC, and it is the main reason I'm storing maxAge.

src/Cookie.php
armpogart: Again, see my other answer. I just made a way to build a minimal valid cookie and am setting defaults in constructor with flag which is on by default.

src/CookieCollection.php
armpogart: What do you mean by not used? `CookieCoollection` implements `Countable`, it uses `ArrayIterator` in `getIterator` metho and also uses `Closure` in `exists` method.

`DateTimeImmutable` and `Psr\Http\Message\ServerRequestInterface` are not used, I just forgot to remove `use` statements. I still have some pending changes, and will remove them either.

src/CookieCollection.php
armpogart: see my other answer.

src/CookieCollection.php
armpogart: Yep, will push the pending changes soon, they are almost ready. The class will have `addToResponse` and `setToResponse`, and it also already have `fromResponse`. So this is just a collection of cookies and easy way to handle (add, modify,...) all the cookies together when needed.

src/CookieCollection.php
armpogart: Yep, I'm working rewriting some of the php doc blocks with more clear texts. And also adding php docs where they are absent in both classes. I'm not a native English speaker, and don't have sufficient experience writing documentation, so pardon any mistakes there.

src/Cookie.php
armpogart: Yep, thanks, will push those changes soon.

src/Cookie.php
armpogart: @samdark I have another idea here. I'm generally against using flags anywhere (constructor, methods...). So what about if I add all available cookie attributes as constructor parameters with their defaults already set to the ones that we consider safe defaults. This will allow to more correct object construction without any flag and will allow also to construct an object with desired attributes without a need to use method chaining for building your desired object in user code, and those methods will be used when needed to modify (actually getting new object, as the class is immutable). So something like this:

```php
public function __construct(
    string $name,
    string $value = '',
    ?DateTimeInterface $expires = null,
    ?int $maxAge = null,
    ?string $domain = null,
    ?string $path = '/',
    ?bool $secure = true,
    ?string $httpOnly = true,
    ?string $sameSite = self::SAME_SITE_LAX
) {
...
}
```

src/Cookie.php
samdark: Technically it can have different expiry and maxAge but it doesn't make sense to set it like that.

> What about cookies that are not set by yii backend, cookies that we just read

We can behave as RFC is suggesting: maxAge will have precedence.

src/Cookie.php
samdark: How about making constructor private and adding two static factories: `Cookie::create()` and `Cookie::createUnsafe()`?

src/CookieCollection.php
samdark: Yes, I've meant only the lines I've commented on.

src/CookieCollection.php
samdark: Me neither. Eventually natives may correct us. That's fine.

src/Cookie.php
armpogart: @samdark Yes, that would be better, I will go that route, but I'm not convinced on naming second factory `Cookie::createUnsafe()` as technically it is just some proposed/recommended defaults and not only about safety. What about naming the second one `Cookie::createWithAttributes` or `Cookie::createWithValues` and it will accept all arguments (cookie attributes) as seen in my constructor example and users who just need custom cookie will use that.

samdark: Your idea about constructor-level defaults looks good. Let's have it.

src/Cookie.php
armpogart: @samdark It can make sense in case if your custom client implementation somehow depends on that (not sure why).
So you propose just to set `Max-Age` anytime the expiry is set? And to also set `Expiry` to the value of  `Max-Age` when parsing the cookie from response and only `Max-Age` is there?

armpogart: Also take into account that `withMaxAge` method is still useful, as it allows to just set expiry in seconds instead of creating `DateTimeInterface`, it can be renamed to `validFor` as it was before, but we will not have a way to set `Max-Age` explicitly if we remove `$maxAge` property.

samdark: I don't think there's a need to make a difference between these two so:

1. Keep one value as private property.
2. Keep two methods of setting it.
3. Add both headers to response.
4. Parse incoming request according to RFC i.e. if Max-Age present, use it. Otherwise, use Expiry.

armpogart: My current `maxAge` (will rename it to `withMaxAge`) method receives `DateInterval` type argument, I still would like to have a method that can also accept integer argument (seconds until expiry) as it can be handy sometimes, but I don't want to add another method for that only and also I see a use for `DateInterval` either (as it is now). The union types would (type hinting) would solve that problem, but unfortunately that RFC will land in PHP 8 by the end of year only.

Do you think we need another method (if we keep `withMaxAge` at all), or keep just this signature until union types and will update (BC break) in some future version?

armpogart: or make `withMaxAge` receive integer instead of `DateInterval` for now to be consistent `Max-Age` cookie attribute that is integer?

P.S. sorry for comment spamming.

armpogart:
Also what about naming of methods in `Cookie` class?

Do we keep them as they are `expires()`, `domain()`, `path()`... or we rename it to `withExpires()`, `withDomain()`, `withPath`... to clearly indicate that they are copying the original object?

src/Cookie.php
samdark: Interval is enough, I think. It is convenient to use it even with seconds:

```php
$seconds = 7200;

$interval = new DateInterval('PT' . $seconds . 'S');
// or
$interval = DateInterval::createFromDateString($seconds . ' seconds');
```

samdark:
Previously cookie was used as write-only so there was no confusion about what these methods do. Now, as you're adding collections, it's a bit different and `with` may come in handy.

armpogart:
Pushed all the changes with tests. Added all getters to `Cookie` for consistency. Removed `expire()` method from `Cookie` as it was used only in one place in `CookieCollection` which is just replaced with setting date in the past. Also I'm in doubt whether `expireWhenBrowserClosed` method needs to remain (though if we remove it, the only way to convert already existing cookie to session cookie would be to create a new one with all the same parameters except expires).

src/CookieCollection.php
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Returns whether the collection already contains the cookie.
+     *
+     * @param Cookie $cookie the cookie to check for
+     * @return bool whether cookie exists
+     * @see has()
+     */
+    public function contains(Cookie $cookie): bool
```
samdark, line 220: Isn't it the same as `has()`?

src/CookieCollection.php
```diff
@@ -0,0 +1,352 @@
...
+        return in_array($cookie, $this->cookies, true);
+    }
+
+    /**
+     * Tests for the existence of the cookie that satisfies the given predicate.
+     *
+     * @param Closure $p The predicate.
+     * @return bool whether the predicate is true for at least on cookie.
+     */
+    public function exists(Closure $p): bool
```
samdark, line 231: What's the use case?

src/CookieCollection.php
```diff
@@ -0,0 +1,352 @@
...
+    {
+        return array_keys($this->cookies);
+    }
+
+    /**
+     * Gets all cookies of the collection as an indexed array.
+     *
+     * @return array The cookie in the collection, in the order they appear in the collection.
+     */
+    public function getCookies(): array
```
samdark, line 257: What's the use case for it?

src/CookieCollection.php
```diff
@@ -0,0 +1,352 @@
...
+    {
+        return array_values($this->cookies);
+    }
+
+    /**
+     * Checks whether the collection is empty (contains no cookies).
+     *
+     * @return bool whether the collection is empty.
+     */
+    public function isEmpty(): bool
```
samdark, line 267: What's the use case for it?

samdark:
I've re-added expire() to Cookie and removed it from collection.

xepozz:
Would be great to have a self-sufficient Collection, not CookieCollection only. 
We have yiisoft/yii2-collection. Would you like to try to port it?

samdark:
I think @romkatsu said he'll do it.

armpogart:
@samdark @xepozz The general idea with `CookieCollection` is to implement some `Collection` interface. which I was planning to extract later on to `yiisoft/arrays`. I didn't even know about existence of `yiisoft/yii2-collection` all these years working with yii2. Is there any reason to have a separate `yiisoft/collection` package for just a couple of classes and interface (it can easily fit to `yiisoft/arrays`). The `Collection` class from `yiisoft/yii2-collection` is just my planned `ArrayCollection` that implements `Collection` interface. `CookieCollection` is specialized and type-hinted case of that collection just for working with cookies. My collection interface was inspired from `CookieCollection` from yii2 and `Collection` from Doctrine collections. I will be glad to extract `Collection` interface after this PR and `ArrayCollection` implementation.

armpogart:
> I've re-added expire() to Cookie and removed it from collection. See #235

Do you mean to re-add `Cookie::expire` method for the mentioned PR. What about keeping also `CookieCollection::expire` in that case as it will save a few lines of code. Without it to expire just one cookie from the collection you would do:
```php
$cookie = $collection->get('someName');
$cookie->expire();
```
with it you would just do:
```php
$collection->expire('someName`);
```

As it is just a helper also in `Cookie` class, it can stay also `CookieCollection`, I just think it can be renamed `expireCookie` for clarity.

src/CookieCollection.php
armpogart: It's just a an implementation of the planned `Collection` interface (see my answer below).

src/CookieCollection.php
armpogart: It's just an implementation of the planned `Collection` interface (see my answer below). Though it would need to be renamed to `getValues` and `getNames` will be renamed to `getKeys`.

src/CookieCollection.php
armpogart: To check existence of the cookie by some attribute, e.g.:
```php
$collection->exists(fn (string $name, Cookie $cookie) => $cookie->getValue() === 'something');
```

Also an implementation of the planned interface (see answer below).

src/CookieCollection.php
armpogart: Nope, `has()` just checks the existence by cookie name, `contains` checks the existence by value (or in this case by reference).

Also implementation of the planned interface (see answer below).

samdark:
> What about keeping also CookieCollection::expire in that case as it will save a few lines of code. 

It doesn't save much:

```php
// collection
$collection->expire('someName');

// cookie
$collection->get('someName')->expire();
```

armpogart:
> > What about keeping also CookieCollection::expire in that case as it will save a few lines of code.
> 
> It doesn't save much:
> 
> ```
> // collection
> $collection->expireCookie('someName');
> 
> // cookie
> $collection->get('someName')->expire();
> ```

This will not expire the cookie in the collection as `Cookie` is immutable, it will create a new cookie. My example is also incorrect. So to `expire` cookie in the collection we would need:
```php
$collection->set('name', $collection->get('name')->expire());
```
Or we will need to re-add `expire()`.

armpogart:
I have pushed some more changes:
- Fixed StyleCI issue.
- Fixed Scrunitizer issues.
 - Added more tests to verify the correct behavior of `CookieCollection::fromArray()` method.
- Renamed some methods and added `CookieCollection::walk()` method which helps to apply user defined function (closure) to each cookie. The usage is for batch editing cookies (later will be used when signing and/or encrypting cookies). The name of the method was chosen to be consistent with native `array_walk` and is planned in `Collection` interface.

samdark:
It fails tests now: https://travis-ci.com/github/yiisoft/yii-web/builds/161267353#L340

samdark:
About the expire... let's re-add the method to collection but keep the method in the Cookie as well.

samdark:
> Renamed some methods and added CookieCollection::walk() method which helps to apply user defined function (closure) to each cookie.

I'd keep collection interface to a minimum for now. @romkatsu is on general collection implementation that is reusable checking Yii 1 collections, yii2-collection and yiisoft/data to get the max out of it. Then it could be used for RBAC, this collection etc. But before it's done we're fine with current implementation.

armpogart:
> > Renamed some methods and added CookieCollection::walk() method which helps to apply user defined function (closure) to each cookie.
> 
> I'd keep collection interface to a minimum for now. @romkatsu is on general collection implementation that is reusable checking Yii 1 collections, yii2-collection and yiisoft/data to get the max out of it. Then it could be used for RBAC, this collection etc. But before it's done we're fine with current implementation.

Ok, I will keep an eye on that and will propose any ideas/changes later, when @romkatsu will publish the work.

I will add an `expire()` method back to `CookieCollection` to and fix the tests today.

samdark (approved):

samdark:
Good job on this one 👍

Issue #248, opened 2020-04-29
error cookie generate


terabytesoftw:
![image](https://user-images.githubusercontent.com/42547589/80642054-d3d2e380-8a33-11ea-9cc3-29fc6f32a5a1.png)

When generating the cookie, the regex validation fails.

samdark:
Looks like #244. @armpogart would you please take a look?

armpogart:
@terabytesoftw Your cookie value contains invalid characters, specifically `,`(comma). Clients (browser) does not allow comma in cookie value. To overcome that limitation you can simply `urlencode()` your value before passing to `Cookie` constructor. The value will be decoded automatically when parsed from response.

@samdark I have chosen not to `urlencode` cookie value automatically (it was `urlencoded` previously, so that any value was allowed) as in majority cases cookie values do not contain invalid characters and can be passed without any encoding (so without overhead of `urlencode` and also less letters/bytes to pass). But I have documented that in `value` property phpdoc, for further documentation. I am not sure where from OP got that cookie value.

armpogart:
@terabytesoftw One extra question to find out if there is a need for any change/fix.

Where did you get that value from? Did you create any custom cookie or is it maybe from session (PHPSESSID).

@samdark I think it may be from PHPSESSID, if that is the case, I will send a PR with the fix to `urlencode` session id, I have checked php source and php actually allows `,`(comma) in session id, so I think it just automatically `urlencode()`s it, and as we use custom `Session` class which creates a cookie without encoding we encounter this error.

terabytesoftw:
Yes PHPSESSID

samdark:
https://github.com/yiisoft/yii-web/pull/249

PR #249, opened 2020-04-30
Use 5 bits per character for sid, to have valid cookie value

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #248 

Not sure why 6 bits per character was used for `sid_bits_per_character` configuration for the session. As pointed in [php documentation](https://www.php.net/manual/en/session.configuration.php#ini.session.sid-bits-per-character) 5 bits is recommended value for most environments. Anyways 6 bits per character was introducing new characters which are invalid for the cookie value, by making session id a bit more stronger.

Another way to fix this issue is to just `urlencode()` session id in `SessionMiddleware`, but I prefer not to introduce additional overhead there.

P.S. I'm planning to review `Session` class and session handling a little bit later on, and will make sure to have a way to use 6 bits per character if the user wants (but will keep 5 as default as recommended by php).

### DIFF
--- a/src/Session/Session.php
+++ b/src/Session/Session.php
@@ -13,7 +13,7 @@ class Session implements SessionInterface
         'use_only_cookies' => 1,
         'cookie_httponly' => 1,
         'use_strict_mode' => 1,
-        'sid_bits_per_character' => 6,
+        'sid_bits_per_character' => 5,
         'sid_length' => 48,
         'cache_limiter' => 'nocache',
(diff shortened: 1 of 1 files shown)


samdark:
5 bits should be enough. 6 was used just in case...

samdark:
Thanks!

PR #254, opened 2020-05-17
Fix PHP 8

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | -

### DIFF
--- a/src/Config/EventConfigurator.php
+++ b/src/Config/EventConfigurator.php
@@ -27,11 +27,11 @@ class EventConfigurator extends AbstractProviderConfigurator
             }
 
             if (!is_array($listeners)) {
-                $type = is_callable($listeners) ? 'callable' : gettype($listeners);
+                $type = $this->isCallable($listeners) ? 'callable' : gettype($listeners);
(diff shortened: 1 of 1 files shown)


src/Config/EventConfigurator.php
```diff
@@ -47,4 +47,13 @@ public function registerListeners(array $eventsListeners): void
...
         }
     }
+
+    private function isCallable($definition): bool
+    {
+        if (is_callable($definition)) {
+            return true;
+        }
+
+        return count($definition) === 2 && in_array($definition[1], get_class_methods($definition[0]) ?? [], true);
```
roxblnfk, line 57: Why do you not check that $definition is an array?

samdark: Done.

roxblnfk: https://github.com/yiisoft/yii-web/commit/61be1117b9d4a2dfa6b793b1af297896198248fd#r39247634

PR #255, opened 2020-05-21
Unit tests patch

hstrychalski:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #105

### DIFF
--- a/tests/Data/Formatter/XmlDataResponseFormatterTest.php
+++ b/tests/Data/Formatter/XmlDataResponseFormatterTest.php
@@ -92,4 +92,58 @@ class XmlDataResponseFormatterTest extends TestCase
         );
         $this->assertSame(['application/xml; UTF-8'], $result->getHeader('Content-Type'));
     }
+
+    public function testFormatterWithObjectTags(): void
+    {
+        $data = new \stdClass();
+        $data->attribute = 'test';
+
+        $factory = new Psr17Factory();
+        $dataResponse = new DataResponse($data, 200, '', $factory);
+        $formatter = new XmlDataResponseFormatter();
+        $formatter = $formatter->withUseObjectTags(true);
+        $result = $formatter->format($dataResponse);
+        $result->getBody()->rewind();
+
+        $this->assertSame(
+            "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<response><stdClass><attribute>test</attribute></stdClass></response>\n",
+            $result->getBody()->getContents()
+        );
+        $this->assertSame(['application/xml; UTF-8'], $result->getHeader('Content-Type'));
+    }
+
+    public function testFormatterWithoutObjectTags(): void
+    {
+        $data = new \stdClass();
+        $data->attribute = 'test';
+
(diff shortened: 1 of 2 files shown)


tests/Data/Formatter/XmlDataResponseFormatterTest.php
```diff
@@ -39,4 +39,111 @@ public function testFormatterEncoding(): void
...
+        $this->assertSame(
+            "<?xml version=\"1.1\" encoding=\"UTF-8\"?>\n<response>test</response>\n",
+            $result->getBody()->getContents()
+        );
+        $this->assertSame(['application/xml; UTF-8'], $result->getHeader('Content-Type'));
+    }
+
+    public function testFormatterRootTag(): void
+    {
+        $factory = new Psr17Factory();
```
xepozz, line 61: Please use mock instead

tests/Data/Formatter/XmlDataResponseFormatterTest.php
```diff
@@ -39,4 +39,111 @@ public function testFormatterEncoding(): void
...
+        $dataResponse = new DataResponse('test', 200, '', $factory);
+        $formatter = new XmlDataResponseFormatter();
+        $formatter = $formatter->withRootTag('exampleRootTag');
+        $result = $formatter->format($dataResponse);
```
xepozz, line 65: You can use fluent interface to configure the `Formatter`:
```php
$result = (new XmlDataResponseFormatter())
                 ->withRootTag('tag')
                 ->format($dataResponse)
```

tests/Data/Formatter/XmlDataResponseFormatterTest.php
```diff
@@ -39,4 +39,111 @@ public function testFormatterEncoding(): void
...
     }
+
+    public function testFormatterVersion(): void
+    {
+        $factory = new Psr17Factory();
+        $dataResponse = new DataResponse('test', 200, '', $factory);
+        $formatter = new XmlDataResponseFormatter();
+        $formatter = $formatter->withVersion('1.1');
+        $result = $formatter->format($dataResponse);
+        $result->getBody()->rewind();
```
xepozz, line 50: Is it really needed?

xepozz (changes requested):

tests/Data/Formatter/XmlDataResponseFormatterTest.php
samdark: Yes. See https://github.com/php-fig/http-message/blob/master/src/StreamInterface.php#L137.

tests/Data/Formatter/XmlDataResponseFormatterTest.php
samdark: Why? We're testing output so we need a working stream to capture it.

samdark (approved):

rustamwin (approved):

samdark:
Thank you!

PR #256, opened 2020-05-24
Fix emitter with not readable stream

roxblnfk:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️


samdark (approved):

samdark:
Thanks!

tests/Emitter/SapiEmitterTest.php
```diff
@@ -1,20 +1,20 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Yiisoft\Yii\Web\Tests\Emitter;
 
-include 'httpFunctionMocks.php';
+include 'Support/httpFunctionMocks.php';
```
xepozz, line 7: @roxblnfk why is it needed?

roxblnfk: Because i moved this file

PR #267, opened 2020-07-06
added "declare (strict_types = 1);" and classes are set as final

thenotsoft:
added "declare (strict_types = 1);", classes are set as final, and bugs found are fixed

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |

### DIFF
--- a/src/Data/DataResponse.php
+++ b/src/Data/DataResponse.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Yiisoft\Yii\Web\Data;
 
 use Psr\Http\Message\ResponseFactoryInterface;
@@ -12,7 +14,7 @@ use Psr\Http\Message\StreamInterface;
  * For example, `['name' => 'Dmitriy']` to be formatted to JSON using {@see \Yiisoft\Yii\Web\Data\Formatter\JsonDataResponseFormatter}
  * when {@see DataResponse::getBody()} is called.
  */
-class DataResponse implements ResponseInterface
+final class DataResponse implements ResponseInterface
 {
     private ResponseInterface $response;
 
--- a/src/Middleware/Csrf.php
+++ b/src/Middleware/Csrf.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Yiisoft\Yii\Web\Middleware;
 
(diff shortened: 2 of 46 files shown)


src/Data/DataResponse.php
```diff
@@ -12,7 +14,7 @@
  * For example, `['name' => 'Dmitriy']` to be formatted to JSON using {@see \Yiisoft\Yii\Web\Data\Formatter\JsonDataResponseFormatter}
  * when {@see DataResponse::getBody()} is called.
  */
-class DataResponse implements ResponseInterface
+final class DataResponse implements ResponseInterface
```
roxblnfk, line 17: I think that here @yiiliveext left the opportunity to extend the decorator

thenotsoft: > I think that here @yiiliveext left the opportunity to extend this decorator

We will remove "final" when necessary.

src/Middleware/Csrf.php
```diff
@@ -91,6 +93,6 @@ private function getTokenFromRequest(ServerRequestInterface $request): ?string
             $token = \reset($headers);
         }
 
-        return TokenMask::remove($token);
+        return is_string($token) ? TokenMask::remove($token) : null;
```
samdark, line 96: In what case it won't be a string?

src/Middleware/HttpCache.php
```diff
@@ -96,16 +98,12 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
         }
 
         $cacheIsValid = $this->validateCache($request, $lastModified, $etag);
+        $response = $handler->handle($request);
+
         if ($cacheIsValid) {
-            return $this->responseFactory
```
samdark, line 100: @rustamwin would you please verify that the change is correct?

src/Middleware/Csrf.php
thenotsoft: `reset ` return boolean if the array is empty

src/Middleware/HttpCache.php
rustamwin: That is correct, yes.

samdark:
👍

PR #269, opened 2020-07-07
Added "JsonBodyParser" middleware

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/Middleware/JsonBodyParser.php
+++ b/src/Middleware/JsonBodyParser.php
@@ -76,17 +76,18 @@ final class JsonBodyParser implements MiddlewareInterface
     /**
      * @return array|object|null
      */
-    private function parse(string $body)
+    private function parse(string $rawBody)
     {
-        try {
-            $result = \json_decode($body, $this->assoc, $this->depth, $this->options);
-            if (\is_array($result) || \is_object($result)) {
-                return $result;
-            }
-        } catch (\JsonException $e) {
-            if ($this->throwException) {
-                throw $e;
-            }
+        $result = \json_decode(
+            $rawBody,
+            $this->assoc,
+            $this->depth,
+            $this->throwException
+                ? $this->options
+                : $this->options & ~JSON_THROW_ON_ERROR
+        );
+        if (\is_array($result) || \is_object($result)) {
+            return $result;
         }
         return null;
     }
--- a/src/Middleware/JsonBodyParser.php
+++ b/src/Middleware/JsonBodyParser.php
@@ -16,7 +16,6 @@ final class JsonBodyParser implements MiddlewareInterface
     private bool $assoc = true;
     private int $depth = 512;
     private int $options = self::DEFAULT_FLAGS;
-    private bool $throwException = true;
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
@@ -62,14 +61,14 @@ final class JsonBodyParser implements MiddlewareInterface
     public function withThrowException(): self
     {
         $new = clone $this;
-        $new->throwException = true;
+        $new->options |= JSON_THROW_ON_ERROR;
         return $new;
     }
 
     public function withoutThrowException(): self
     {
         $new = clone $this;
-        $new->throwException = false;
+        $new->options &= ~JSON_THROW_ON_ERROR;
         return $new;
     }
 
@@ -78,14 +77,7 @@ final class JsonBodyParser implements MiddlewareInterface
      */
     private function parse(string $rawBody)
     {
-        $result = \json_decode(
-            $rawBody,
-            $this->assoc,
-            $this->depth,
-            $this->throwException
-                ? $this->options
-                : $this->options & ~JSON_THROW_ON_ERROR
-        );
+        $result = \json_decode($rawBody, $this->assoc, $this->depth, $this->options);
         if (\is_array($result) || \is_object($result)) {
             return $result;
         }
--- a/src/Middleware/JsonBodyParser.php
+++ b/src/Middleware/JsonBodyParser.php
@@ -49,7 +49,7 @@ final class JsonBodyParser implements MiddlewareInterface
     {
         try {
             $result = json_decode($body, $this->assoc, 512, JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_IGNORE);
-            if (($this->assoc === true && is_array($result)) || ($this->assoc === false && is_object($result))) {
+            if (is_array($result) || is_object($result)) {
                 return $result;
             }
         } catch (\JsonException $e) {
--- a/src/Middleware/JsonBodyParser.php
+++ b/src/Middleware/JsonBodyParser.php
@@ -21,7 +21,7 @@ final class JsonBodyParser implements MiddlewareInterface
     {
         $contentType = $request->getHeaderLine(Header::CONTENT_TYPE);
 
-        if ($contentType && strpos(strtolower($contentType), 'application/json') !== false) {
+        if ($contentType && \strpos(\strtolower($contentType), 'application/json') !== false) {
             $request = $request->withParsedBody(
                 $this->parse($request->getBody()->getContents())
             );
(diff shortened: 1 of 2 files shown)


src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,48 @@
...
+        if ($contentType && strpos(strtolower($contentType), 'application/json') !== false) {
+            $request = $request->withParsedBody(
+                $this->parse($request->getBody()->getContents())
+            );
+        }
+
+        return $handler->handle($request);
+    }
+
+    public function throwException(bool $value): self
```
roxblnfk, line 29: Immutale methods should by named with `with` prefix

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,48 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class JsonBodyParser implements MiddlewareInterface
+{
+    private bool $throwException = true;
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $contentType = $request->getHeaderLine('Content-Type');
```
roxblnfk, line 18:
```suggestion
        $contentType = $request->getHeaderLine(\Yiisoft\Http\Header::CONTENT_TYPE);
```

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,48 @@
...
+        $new = clone $this;
+        $new->throwException = $value;
+        return $new;
+    }
+
+    private function parse(string $body): array
+    {
+        try {
+            $result = json_decode($body, true, 512, JSON_THROW_ON_ERROR);
+            return is_array($result) ? $result : [];
```
roxblnfk, line 40: Why default is `[]` (why not `null`)? The `withParsedBody` method supports `null|array|object`.

thenotsoft: > Why default value is empty array (why not `null`)? The `withParsedBody` method supports `null|array|object`.

an array is expected when parsing json and we return array

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,78 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Http\Header;
+
+final class JsonBodyParser implements MiddlewareInterface
+{
+    private bool $assoc = true;
+    private int $depth = 512;
+    private int $options = JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_IGNORE;
```
roxblnfk, line 17:
```suggestion
{
    private const DEFAULT_FLAGS = JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_IGNORE;
    private bool $assoc = true;
    private int $depth = 512;
    private int $options = self::DEFAULT_FLAGS;
```

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $new = clone $this;
+        $new->depth = $value;
+        return $new;
+    }
+
+    public function withOptions(int $value): self
+    {
+        $new = clone $this;
+        $new->options = $value;
```
roxblnfk, line 50:
```suggestion
        $new->options = self::DEFAULT_FLAGS | $value;
```

roxblnfk (commented):
What you think about this?

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function withThrowException(bool $value): self
+    {
+        $new = clone $this;
+        $new->throwException = $value;
+        return $new;
+    }
+
+    /**
+     * @return mixed
```
xepozz, line 63:
```suggestion
     * @return array[]|object|null
```

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    private function parse(string $body)
+    {
+        try {
+            $result = json_decode($body, $this->assoc, $this->depth, $this->options);
+            if (is_array($result) || is_object($result)) {
+                return $result;
+            }
+        } catch (\JsonException $e) {
+            if ($this->throwException) {
```
xepozz, line 73: How about to log this error?

tests/Middleware/JsonBodyParserTest.php
```diff
@@ -0,0 +1,167 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Yiisoft\Yii\Web\Middleware\JsonBodyParser;
+
+final class JsonBodyParserTest extends TestCase
+{
+    public function testProcess()
+    {
+        $request = $this->createMock(ServerRequestInterface::class);
```
xepozz, line 17: Please extract configuring of request to separated method and reuse it

xepozz (changes requested):

src/Middleware/JsonBodyParser.php
roxblnfk: `array|object|null`

src/Middleware/JsonBodyParser.php
```diff
@@ -64,15 +63,9 @@ public function withThrowException(bool $value): self
...
-            }
-        } catch (\JsonException $e) {
-            if ($this->throwException) {
-                throw new \InvalidArgumentException('Invalid JSON data in request body: ' . $e->getMessage());
-            }
+        $result = json_decode($body, $this->assoc, $this->depth, $this->options);
+        if (is_array($result) || is_object($result)) {
+            return $result;
         }
         return null;
```
xepozz:
```suggestion
        if (false === $result) {
            return null;
        }
        return $result;
```

thenotsoft: With such a check, this will pass https://3v4l.org/9e159
`withParsedBody` accepts only `array|object|null`

src/Middleware/JsonBodyParser.php
xepozz: Sure

src/Middleware/JsonBodyParser.php
xepozz: https://tools.ietf.org/html/rfc7159#section-13
Strings, numbers and booleans are valid values.
Use https://www.php.net/manual/ru/function.json-last-error.php to detect the decoding error

thenotsoft: > https://tools.ietf.org/html/rfc7159#section-13
> Strings, numbers and booleans are valid values.
> Use https://www.php.net/manual/ru/function.json-last-error.php to detect the decoding error

but `withParsedBody` not supported other data type

xepozz (approved):

roxblnfk (approved):

src/Middleware/JsonBodyParser.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->assoc = $assoc;
+        $this->depth = $depth;
+        $this->options = $options;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $contentType = $request->getHeaderLine(Header::CONTENT_TYPE);
+
+        if ($contentType && \strpos(\strtolower($contentType), 'application/json') !== false) {
```
romkatsu, line 33: maybe it's better to put it in the method?

```php
        if ($this->isJsonRequest($request)) {
            $request = $request->withParsedBody(
                $this->parse($request->getBody()->getContents())
            );
        }
```

```php
    private function isJsonRequest(ServerRequestInterface $request): bool
    {
        $contentType = $request->getHeaderLine(Header::CONTENT_TYPE);

        return $contentType && stripos($contentType, 'application/json') !== false;
    }
```

romkatsu (changes requested):

romkatsu (approved):

xepozz (approved):

thenotsoft:
absorbed by https://github.com/yiisoft/request-body-parsers/pull/1

PR #282, opened 2020-07-14
Fix using event configurator with DI references

viktorprogger:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | ❌

Fixing usage with aliases in DI configuration:
`['someAlias' => ClassName::class]` in DI configuration and `['someALias', 'methodName']` in `events-console.php`.

### DIFF
--- a/src/Config/EventConfigurator.php
+++ b/src/Config/EventConfigurator.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace Yiisoft\Yii\Web\Config;
 
+use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\ContainerInterface;
+use Psr\Container\NotFoundExceptionInterface;
 use Yiisoft\EventDispatcher\Provider\AbstractProviderConfigurator;
 use Yiisoft\EventDispatcher\Provider\Provider;
 use Yiisoft\Injector\Injector;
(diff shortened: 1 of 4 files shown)


src/Config/EventConfigurator.php
```diff
@@ -23,23 +25,33 @@ public function __construct(Provider $listenerProvider, ContainerInterface $cont
...
+     * @param array $eventListeners Event listener list in format ['eventName1' => [$listener1, $listener2, ...]]
      */
-    public function registerListeners(array $eventsListeners): void
+    public function registerListeners(array $eventListeners): void
     {
-        foreach ($eventsListeners as $eventName => $listeners) {
+        foreach ($eventListeners as $eventName => $listeners) {
             if (!is_string($eventName)) {
-                throw new \RuntimeException('Incorrect event listener format. Format with event name must be used.');
+                $message = 'Incorrect event listener format. Format with event name must be used.';
```
xepozz, line 35: Would be nice to inline into creating object

src/Config/EventConfigurator.php
```diff
@@ -59,6 +71,23 @@ private function isCallable($definition): bool
...
+            && array_keys($definition) === [0, 1]
+            && is_string($definition[0])
+            && $this->container->has($definition[0])
+        ) {
+            try {
+                $object = $this->container->get($definition[0]);
+
+                return method_exists($object, $definition[1]);
+            } catch (NotFoundExceptionInterface|ContainerExceptionInterface $exception) {
+                $message = "Could not instantiate event listener or listener class has invalid configuration.";
```
thenotsoft, line 85:
```suggestion
       throw new InvalidListenerConfigurationException(
            "Could not instantiate event listener or listener class has invalid configuration.",
            0,
            $exception
        );
```

thenotsoft, line 85: It might be better to remove $message and pass the message direct to Exception ?

src/Config/EventConfigurator.php
```diff
@@ -23,23 +25,42 @@
...
-                    throw new \RuntimeException("Listener must be a callable. $type given.");
+                try {
+                    if (!$this->isCallable($callable)) {
+                        $type = gettype($listeners);
+
+                        throw new InvalidListenerConfigurationException(
+                            "Listener must be a callable. $type given."
+                        );
+                    }
+                } catch (NotFoundExceptionInterface|ContainerExceptionInterface $exception) {
```
roxblnfk, line 57: NotFoundExceptionInterface extends ContainerExceptionInterface

src/Config/EventConfigurator.php
```diff
@@ -59,6 +80,17 @@
...
         }
 
-        return is_array($definition) && array_keys($definition) === [0, 1] && in_array($definition[1], get_class_methods($definition[0]) ?? [], true);
+        if (
+            is_array($definition)
+            && array_keys($definition) === [0, 1]
+            && is_string($definition[0])
+            && $this->container->has($definition[0])
+        ) {
+            $object = $this->container->get($definition[0]);
```
roxblnfk, line 89: I think this method should not throw an exception

viktorprogger: This method does not throw an exception outside its object. These exceptions are catched above and chained into a more suitable exception.

samdark:
👍

PR #295, opened 2020-08-06
Add withOnlyRenderer method

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | add withOnlyRenderer method

Sometimes it is more convenient to use withOnlyRenderer instead of withoutRenderers() && withRenderer().

A single render is needed when the output should be the same regardless of the request.


samdark:
👍

PR #296, opened 2020-08-06
Allow set return content type

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | Allow set return content type

A single render is needed when the output should be the same regardless of the request.


samdark:
1. What's the use-case?
2. Can it be done as ->withoutRenderers()->withRenderer()?

romkatsu:
>     1. What's the use-case?
> 
>     2. Can it be done as ->withoutRenderers()->withRenderer()?

1. rest api - regardless of the request you need to return errors in json format
2. No. These methods only add-remove the Renderers. The desired render is taken from the accept header

samdark:
👍

src/ErrorHandler/ErrorCatcher.php
```diff
@@ -112,10 +112,6 @@ private function getRenderer(string $contentType): ?ThrowableRendererInterface
 
     private function getContentType(ServerRequestInterface $request): string
     {
-        if ($this->contentType !== null) {
-            return $this->contentType;
-        }
-
         try {
             foreach (HeaderHelper::getSortedAcceptTypes($request->getHeader('accept')) as $header) {
```
roxblnfk:
```suggestion
            foreach (HeaderHelper::getSortedAcceptTypes($request->getHeader(Header::ACCEPT)) as $header) {
```

Issue #275, opened 2020-07-07
Separate package


xepozz:
If you look at this package, you can find components that can be used outside the framework and the `yii-web` package.

I studied the question a bit and suggest the following actions:
1. <strike>Take out the cookie package.
    Classes for working with cookies are self-sufficient and can be reused in other projects. The only "parasitic" dependency is the `ResponseInterface` in the `CookieCollection`, but that’s not so scary.</strike>
2. <strike>Combine the `Session` and `Flash` components and put them into one package (`session`?)
    These two components are also self-sufficient and independent, and also, although this is hardly believed, can be reused outside the framework.</strike>
3. `Middlewares` folder
4. <strike>It would be nice to render `ErrorHandler` too.
    Now this dependency is needed only in `Application`. This component could potentially be used outside of the `yii-web` package. Moreover, the component also contains templates for rendering errors.</strike>
5. <strike>`Data`
    Response formatting is guaranteed to be used outside the framework. I think there’s no need to even discuss anything here. This is worth putting in a separate package.</strike>

samdark:
@xepozz would that allow to slim down yii-web package so it will become suitable as dependency for API (yii-rest)?

xepozz:
I think yes, it is. Furthermore yii-web can be really micro framework

samdark:
5. https://github.com/yiisoft/data-response

samdark:
Point 3 isn't a good one. Extracting all middleware in a single packages isn't good.

roxblnfk:
About `CookieCollection`. 
What do you think about moving this to the `http` package?

xepozz:
A good idea. I'm all for it.
Maybe move `Session` too?

samdark:
Currently HTTP is mainly framework-independent and almost PSR-independent. The only place where PSR is used is `HeaderHelper::getSortedAcceptTypesFromRequest(RequestInterface $request)` and it likely worth removing this method (see https://github.com/yiisoft/http/issues/11).

We need to decide about the scope of this package:

1. Should it stay independent and represent HTTP RFCs only?
2. May it depend on PSRs? Which ones?
3. May it depend on the framework?

Based on answers we'll decide on cookies and session.

IMO, both cookies and session should either to stay in yii-web or be extracted in their own separate packages.

samdark:
Decided to have separate repos for session / cookie.

rustamwin:
- `User` folder -> [auth](https://github.com/yiisoft/auth)?

[...]

PR #298, opened 2020-08-10
Added AuthInterface implementation for the User.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

`Yiisoft\Auth\AuthInterface` implementation for the `User` class to use it in `Auth` middleware.
I'm not sure on naming and having default `authUrl` predefined.

P.S. Will add tests as soon you give the green light.

### DIFF
--- a/src/User/UserAuth.php
+++ b/src/User/UserAuth.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Yii\Web\User;
+
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Yiisoft\Auth\AuthInterface;
+use Yiisoft\Auth\IdentityInterface;
+use Yiisoft\Http\Status;
+
+final class UserAuth implements AuthInterface
+{
+    private string $authUrl = '/login';
+    private User $user;
+    private ResponseFactoryInterface $responseFactory;
+
+    public function __construct(User $user, ResponseFactoryInterface $responseFactory)
+    {
+        $this->user = $user;
+        $this->responseFactory = $responseFactory;
+    }
+
+    public function authenticate(ServerRequestInterface $request): ?IdentityInterface
+    {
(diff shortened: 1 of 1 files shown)


rustamwin:
https://github.com/yiisoft/yii-web/issues/275#issuecomment-666676216?

armpogart:
I've looked through that issue, but thought it concerns only cookie and session.
If `User` will be moved to `Auth` it will also make the `auth` package to depend on cookie and session packages.

I think `User` is more specific for yii framework and not overall community as a separate package. I'm not sure on usability of `User` outside the framework. User implementation is usually strongly coupled to database entity or the code in other frameworks or frameworkless applications, so I see little value in having it there.
What about `AccessCheckerInterface` in `User` class in that case?

If the decision will be made to move it to `auth` package, this can also be moved with it as it is in the same folder.

samdark:
1. Would that make `AutoLoginMiddleware` duplicate functionality?
2. Should we do something about `User::loginByAccessToken()` then?

armpogart:
> 1. Would that make `AutoLoginMiddleware` duplicate functionality?
> 2. Should we do something about `User::loginByAccessToken()` then?

1. I'm not sure how this is supposed to duplicate `AutLoginMiddleware` functionality. This is `AuthInterface` functionality for `Auth` middleware that will be used on route or group of routes to ensure that the user is already logged in based purely on session. `AutoLoginMiddleware` on the other hand logs in user even if the session is expired based on some cookie. I don't see anything in common.
2. `User::loginByAccessToken` must be removed from the `User` in my opinion. We can have another `IdentityInterface` implementation and middleware specific for that. And we can have another `AuthInterface` implementation specifically for that, as it will be used mainly to authenticate API requests and not general user.

samdark:
Alright. Does it make sense to do point 2 in this pull request?

armpogart:
No, I've drafted all things needed for that in separate issue #304. There are a couple of packages that need to be refined for that (not only yii-web).

samdark (approved):

samdark:
👍

Issue #302, opened 2020-08-16
Error page stacktrace call arguments are inverted


samdark:
![90342026-85127f00-e00d-11ea-9ed4-56031948506c](https://user-images.githubusercontent.com/47294/90343718-cb6eda80-e01b-11ea-8255-629ccb20b375.png)

Found by @roxblnfk

samdark:
Add the following to index action of site controller:


```php
(function () {
    throw new Exception('s');
})([1, 2, 3], true);
```

samdark:
https://github.com/yiisoft/yii-web/pull/303

PR #303, opened 2020-08-16
Fix error handler

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/yii-web/issues/302, not rendering fallback error, not rendering server info

### DIFF
--- a/src/ErrorHandler/ErrorHandler.php
+++ b/src/ErrorHandler/ErrorHandler.php
@@ -67,7 +67,7 @@ final class ErrorHandler
             $this->log($t);
             return $this->exposeDetails ? $renderer->renderVerbose($t) : $renderer->render($t);
         } catch (\Throwable $t) {
-            return nl2br($t);
+            return (string)$t;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/ErrorHandler/ErrorHandler.php
```diff
@@ -67,7 +67,7 @@ public function handleCaughtThrowable(\Throwable $t, ThrowableRendererInterface
             $this->log($t);
             return $this->exposeDetails ? $renderer->renderVerbose($t) : $renderer->render($t);
         } catch (\Throwable $t) {
-            return nl2br((string)$t);
+            return (string)$t;
```
samdark, line 70: This one fixes error rendering fallback. Previously it was just giving out 500 w/o any clues.

src/ErrorHandler/HtmlRenderer.php
```diff
@@ -292,6 +292,8 @@ private function argumentsToString(array $args): string
                 $args[$key] = "<span class=\"number\">$key</span> => $args[$key]";
             }
         }
+
+        ksort($args);
```
samdark, line 296: This one fixes arguments order in stacktrace.

PR #307, opened 2020-09-13
proposal for different middleware dispatching flow

denyadzi:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

Inspired by [laminas/laminas-stratigility middleware pipe](https://github.com/laminas/laminas-stratigility/blob/3.3.x/src/MiddlewarePipe.php). 

It seems to me that there is an overhead with creating the whole handler-wrapped middleware stack before handling the request. Here next handlers are created right before they're passed to the middleware process(). I used SplStack just like SplQueue is used in laminas-stratigility, I don't know whether it is more effective than simple array, but looks little bit cleaner

### DIFF
--- a/src/MiddlewareDispatcher.php
+++ b/src/MiddlewareDispatcher.php
@@ -21,25 +21,18 @@ use Yiisoft\Yii\Web\Event\BeforeMiddleware;
 final class MiddlewareDispatcher
 {
     /**
-     * @var MiddlewareInterface[]
+     * @var \SplStack
      */
-    private array $middlewares = [];
+    private \SplStack $middlewares;
 
     private RequestHandlerInterface $nextHandler;
     private ContainerInterface $container;
-
-    /**
-     * Contains a chain of middleware wrapped in handlers.
-     * Each handler points to the handler of middleware that will be processed next.
-     *
-     * @var RequestHandlerInterface|null stack of middleware
-     */
-    private ?RequestHandlerInterface $stack = null;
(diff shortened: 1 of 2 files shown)


yiiliveext:
> Q	A
> Is bugfix?	
> New feature?	
> Breaks BC?	
> Fixed issues	
> Inspired by [laminas/laminas-stratigility middleware pipe](https://github.com/laminas/laminas-stratigility/blob/3.3.x/src/MiddlewarePipe.php).
> 
> It seems to me that there is an overhead with creating the whole handler-wrapped middleware stack before handling the request. Here next handlers are created right before they're passed to the middleware process(). I used SplStack just like SplQueue is used in laminas-stratigility, I don't know whether it is more effective than simple array, but looks little bit cleaner

What about working in a request loop? (Swool, Roadrunner, etc)

denyadzi:
@yiiliveext Just looked from the perspective of swoole-like environments and realized that the current aproach is better there. So, may I leave only the MiddlewareDispatcherTest improvement for this proposal, checkouting everything else?

samdark:
Yep. If that pass, feel free to leave it.

denyadzi:
Left MiddlewareDispatcherTest additions only, all tests are passing

samdark (approved):

samdark:
👍

PR #319, opened 2020-10-15
Replace phan with psalm

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -


composer.json
```diff
@@ -47,7 +47,7 @@
     },
     "require-dev": {
         "nyholm/psr7": "^1.0",
-        "phan/phan": "^3.0",
+        "vimeo/psalm": "^3.17",
         "phpunit/phpunit": "^9.3",
```
roxblnfk, line 51:
```suggestion
        "phpunit/phpunit": "^9.3",
        "vimeo/psalm": "^3.17",
```

psalm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="6"
```
roxblnfk, line 3: are errors occurring at lower values?

roxblnfk (approved):

psalm.xml
rustamwin: Documentation [says](https://psalm.dev/docs/running_psalm/error_levels/): Level 1 is the most strict, level 8 is the most lenient.

Issue #272, opened 2020-07-07
AutoLoginMiddleware bug


xepozz:
Type of the first parameter of method `\Yiisoft\Yii\Web\User\AutoLogin::addCookie` is `\Yiisoft\Yii\Web\User\AutoLoginIdentityInterface`, but `\Yiisoft\Yii\Web\User\User::getIdentity` returns `\Yiisoft\Auth\IdentityInterface` on `src/User/AutoLoginMiddleware.php:44`

samdark:
No a bug:

```php
interface AutoLoginIdentityInterface extends IdentityInterface
```

xepozz:
I've added testcase see #322

xepozz:
Should be moved to https://github.com/yiisoft/user

PR #322, opened 2020-10-24
Test case for bug #272

xepozz:


xepozz:
Should be moved to https://github.com/yiisoft/user

samdark:
Pull requests can't be moved. Closing it here. Linked from the issue itself.

Issue #309, opened 2020-09-16
Move error handler out of application


samdark:
We need to ensure that configuring of container is wrapped into error handler. Solution is to move error handler out of application into index.php.

samdark:
The problem with it is that error handler is configured via container.

mj4444ru:
Two error handlers are required. The first is easy to handle errors during the configuration creation phase. The second one is configurable to handle errors when the application is running.

samdark:
Yeah. Something like that. Do you want to attempt solving it?

mj4444ru:
I can suggest this option.
  - InternalServerError.php - Production.
  - InternalServerErrorDev.php - Output of debug information.

This code will also handle the exception if something went wrong in the main handler.
The user can freely customize the output by displaying, for example, a static branded page.

```php
if (!is_file($autoload)) {
    die('You need to set up the project dependencies using Composer');
}

require_once $autoload;

set_exception_handler(function(Throwable $exception) {
    $tpl = ini_get('display_errors') === '1' ? 'InternalServerErrorDev.php': 'InternalServerError.php';
    require dirname(__DIR__) . '/resources/' . $tml;
});
// Don't do it in production, assembling takes it's time
Builder::rebuild();
$startTime = microtime(true);
```

samdark:
I wonder if we can create an instance of https://github.com/yiisoft/yii-web/blob/master/src/ErrorHandler/ErrorHandler.php right at the start of the file, call `register()`, init container, and then call `unregister()` after container is initialized so container's handler will take control.

mj4444ru:
On a note

```php
    public function start(): void
    {
        $this->errorHandler->register();           <--- without unregister
        $this->eventDispatcher->dispatch(new ApplicationStartup());
    }

    public function shutdown(): void
    {
        $this->eventDispatcher->dispatch(new ApplicationShutdown());
    }
```

```php
    public function register(): void
    {
        $this->disableDisplayErrors();
        set_exception_handler([$this, 'handleThrowable']);
        set_error_handler([$this, 'handleError']);

        if ($this->memoryReserveSize > 0) {
            $this->memoryReserve = str_repeat('x', $this->memoryReserveSize);
        }
        register_shutdown_function([$this, 'handleFatalError']);           <-- Remains active after unregistrer.
    }

    public function unregister(): void
    {
        restore_error_handler();
        restore_exception_handler();
    }
```

mj4444ru:
[...]

PR #326, opened 2020-10-25
Added handy setters to ErrorHandler

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | 

It's needed for #309

### DIFF
--- a/src/ErrorHandler/ErrorHandler.php
+++ b/src/ErrorHandler/ErrorHandler.php
@@ -4,11 +4,15 @@ declare(strict_types=1);
 
 namespace Yiisoft\Yii\Web\ErrorHandler;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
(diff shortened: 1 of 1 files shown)


src/ErrorHandler/ErrorHandler.php
```diff
@@ -19,7 +23,6 @@ final class ErrorHandler
     private string $memoryReserve = '';
     private bool $exposeDetails = true;
 
-    private LoggerInterface $logger;
     private ThrowableRendererInterface $defaultRenderer;
 
     public function __construct(LoggerInterface $logger, ThrowableRendererInterface $defaultRenderer)
```
samdark, line 28: Shouldn't we remove logger from constructor because it's already in a setter?

xepozz: I don't think so. Then `$logger` be null in processing if `setLogger` won't be called.

PR #330, opened 2020-11-03
Add assets aliases

iamsaint:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


samdark:
That's not correct place to put these aliases since yii-web knows nothing about assets.

PR #332, opened 2020-11-03
Add exception handler middleware

samdark:
Should be useful for mapping domain exceptions to certain HTTP response codes.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


src/Middleware/ExceptionResponder.php
```diff
@@ -0,0 +1,48 @@
...
+use Yiisoft\Injector\Injector;
+
+class ExceptionResponder implements MiddlewareInterface
+{
+    private array $exceptionMap;
+    private ResponseFactoryInterface $responseFactory;
+    private Injector $injector;
+
+    /**
+     * @param array $exceptionMap
```
roxblnfk, line 19:
```suggestion
     * @param int[]|callable[] $exceptionMap Closure must return ResponseInterface
```

src/Middleware/ExceptionResponder.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
```
roxblnfk, line 2:
```suggestion

declare(strict_types=1);

```

src/Middleware/ExceptionResponder.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    private array $exceptionMap;
+    private ResponseFactoryInterface $responseFactory;
+    private Injector $injector;
+
+    /**
+     * @param array $exceptionMap
+     * @psalm-param array{string, int|callable}
+     */
+    public function __construct(array $exceptionMap, ResponseFactoryInterface $responseFactory, Injector $injector)
```
roxblnfk, line 22: You can specify the ContainerInterface in the parameters and get the ResponseFactoryInterface and Injector already in the catch body

samdark: I don't see a profit in doing it. Makes code less clean.

roxblnfk: Mb. I just don't like that you pass the exception map along with the dependencies.

rustamwin (approved):
I like the idea.

samdark:
This one can potentially become part of https://github.com/yiisoft/error-handler

PR #334, opened 2020-11-04
Move out user and clean dependencies

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #275


composer.json
```diff
@@ -30,20 +30,13 @@
...
         "yiisoft/http": "^1.0",
         "yiisoft/injector": "^1.0",
         "yiisoft/network-utilities": "^3.0@dev",
         "yiisoft/router": "^3.0@dev",
-        "yiisoft/security": "^1.0",
-        "yiisoft/serializer": "^3.0@dev",
-        "yiisoft/session": "^3.0@dev",
-        "yiisoft/validator": "^3.0@dev",
-        "yiisoft/yii-event": "^3.0@dev"
+        "yiisoft/validator": "^3.0@dev"
```
samdark, line 39: yiisoft/user should be added

samdark (changes requested):

composer.json
xepozz: Why?

roxblnfk (approved):
I also think there is no need to add a `user` as as dependency

samdark (approved):

samdark:
Finished and merged. Thank you!

Issue #327, opened 2020-10-25
Move ErrorHandler into separated package


xepozz:
After https://github.com/yiisoft/yii-web/pull/325 be merged `ErrorHandler` can be independent package.

samdark:
#325 is merged.

xepozz:
@samdark can you please create a new repo?

samdark:
https://github.com/yiisoft/error-handler

PR #335, opened 2020-11-04
Moved to error-handler

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #327


samdark:
composer.json should be adjusted.

samdark:
I've added the package and fixed some configs but it's still giving

```
  [ErrorException]
  Undefined index: htmlRenderer
```

PR #357, opened 2021-02-28
Fix headers parsing to ServerRequestFactory

devanych:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

This PR fixes the following:

1. Prevents duplication of the `Host` header in the request.
2. Adds parsing of headers prefixed with `REDIRECT_` and `CONTENT_`.
3. Removes empty headers in the request, this is especially true for `Content-Length` and `Content-Type`.
4. Makes the first letter of each word in the header name uppercase when parsing, like the `getallheaders()` function.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Yiisoft\Yii\Web;
 
+use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestFactoryInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\StreamFactoryInterface;
@@ -11,6 +12,23 @@ use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UploadedFileFactoryInterface;
 use Psr\Http\Message\UriFactoryInterface;
 use Psr\Http\Message\UriInterface;
+use RuntimeException;
+
+use function array_filter;
+use function array_key_exists;
+use function explode;
+use function fopen;
+use function function_exists;
+use function getallheaders;
+use function is_array;
+use function is_resource;
+use function is_string;
+use function preg_match;
+use function str_replace;
+use function strncmp;
+use function strtolower;
+use function substr;
+use function ucwords;
 
 final class ServerRequestFactory
 {
@@ -40,7 +58,7 @@ final class ServerRequestFactory
             $_GET,
             $_POST,
             $_FILES,
-            \fopen('php://input', 'rb') ?: null
+            fopen('php://input', 'rb') ?: null
         );
     }
 
@@ -59,14 +77,17 @@ final class ServerRequestFactory
     {
         $method = $server['REQUEST_METHOD'] ?? null;
         if ($method === null) {
-            throw new \RuntimeException('Unable to determine HTTP request method.');
+            throw new RuntimeException('Unable to determine HTTP request method.');
         }
 
         $uri = $this->getUri($server);
-
         $request = $this->serverRequestFactory->createServerRequest($method, $uri, $server);
 
         foreach ($headers as $name => $value) {
+            if ($name === 'Host' && $request->hasHeader('Host')) {
+                continue;
+            }
+
             $request = $request->withAddedHeader($name, $value);
         }
 
@@ -86,12 +107,12 @@ final class ServerRequestFactory
             return $request;
         }
 
-        if (\is_resource($body)) {
+        if (is_resource($body)) {
             $body = $this->streamFactory->createStreamFromResource($body);
-        } elseif (\is_string($body)) {
+        } elseif (is_string($body)) {
             $body = $this->streamFactory->createStream($body);
         } elseif (!$body instanceof StreamInterface) {
-            throw new \InvalidArgumentException('Body parameter for ServerRequestFactory::createFromParameters() must be instance of StreamInterface, resource or null.');
+            throw new InvalidArgumentException('Body parameter for ServerRequestFactory::createFromParameters() must be instance of StreamInterface, resource or null.');
         }
 
         return $request->withBody($body);
@@ -108,7 +129,7 @@ final class ServerRequestFactory
         }
 
         if (isset($server['HTTP_HOST'])) {
-            if (1 === \preg_match('/^(.+):(\d+)$/', $server['HTTP_HOST'], $matches)) {
+            if (1 === preg_match('/^(.+):(\d+)$/', $server['HTTP_HOST'], $matches)) {
                 $uri = $uri->withHost($matches[1])->withPort((int) $matches[2]);
             } else {
                 $uri = $uri->withHost($server['HTTP_HOST']);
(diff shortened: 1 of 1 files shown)


src/ServerRequestFactory.php
```diff
@@ -59,14 +77,17 @@ public function createFromParameters(array $server, array $headers = [], array $
...
+            throw new RuntimeException('Unable to determine HTTP request method.');
         }
 
         $uri = $this->getUri($server);
-
         $request = $this->serverRequestFactory->createServerRequest($method, $uri, $server);
 
         foreach ($headers as $name => $value) {
+            if ($name === 'Host' && $request->hasHeader('Host')) {
+                continue;
```
roxblnfk, line 88:
```suggestion
            if ($name === 'Host') {
                $request->withHeader('Host', $value)
                continue;
```
Do you think, depending on the implementation of the ResponseInterface, could a different result be achieved?
If yes, what would be the best case?

roxblnfk, line 88:
```suggestion
            if ($name === 'Host') {
                $request->withHeader('Host', $value)
                continue;
```
Do you think, depending on the implementation of the ResponseInterface, could a different result be achieved?
If yes, what would be the best case?

src/ServerRequestFactory.php
```diff
@@ -132,29 +153,47 @@ private function getUri(array $server): UriInterface
...
-     */
     private function getHeadersFromGlobals(): array
     {
-        if (\function_exists('\getallheaders')) {
-            $headers = \getallheaders();
-            if ($headers === false) {
-                $headers = [];
+        if (function_exists('getallheaders')) {
+            $headers = getallheaders();
+            return $headers === false ? [] : array_filter($headers, static fn (string $value): bool => $value !== '');
```
roxblnfk, line 160: Is it possible to get a empty header value here?

src/ServerRequestFactory.php
```diff
@@ -132,29 +153,47 @@ private function getUri(array $server): UriInterface
...
+                continue;
             }
-        } else {
-            $headers = [];
-            foreach ($_SERVER as $name => $value) {
-                if (strncmp($name, 'HTTP_', 5) === 0) {
-                    $name = str_replace(' ', '-', strtolower(str_replace('_', ' ', substr($name, 5))));
-                    $headers[$name] = $value;
+
+            if (strncmp($name, 'REDIRECT_', 9) === 0) {
```
roxblnfk, line 170: I am not sure if there is a place for `REDIRECT_` parameters among the headers. What parameter names do you mean?

src/ServerRequestFactory.php
```diff
@@ -132,29 +153,47 @@ private function getUri(array $server): UriInterface
...
                 }
             }
+
+            if (strncmp($name, 'HTTP_', 5) === 0) {
+                $headers[$this->normalizeHeaderName(substr($name, 5))] = $value;
+                continue;
+            }
+
+            if (strncmp($name, 'CONTENT_', 8) === 0) {
+                $headers[$this->normalizeHeaderName($name)] = $value;
```
roxblnfk, line 184: These values can be empty

src/ServerRequestFactory.php
devanych: PSR implementations must add `Host` and make it the first one.
https://tools.ietf.org/html/rfc7230#section-5.4

If we do as you suggest, the header will always be overwritten and `Host` will not be the first one.

devanych: PSR implementations must add `Host` and make it the first one.
https://tools.ietf.org/html/rfc7230#section-5.4

If we do as you suggest, the header will always be overwritten and `Host` will not be the first one.

src/ServerRequestFactory.php
devanych: Yes.

src/ServerRequestFactory.php
devanych: Apache prefixes environment variables with `REDIRECT_` if they are added by rewrite rules. :)

src/ServerRequestFactory.php
devanych: Yes

samdark (approved):

samdark:
👍

src/ServerRequestFactory.php
roxblnfk: These `REDIRECT_` variables are not headers.
I was checked two servers and wath i see:

`REDIRECT_STATUS = 200` (nginx)

and

```
REDIRECT_TZ = Europe/Moscow
REDIRECT_STATUS = 200
```
(nginx + apache)

Are you suggesting converting this to headers? Do not do it this way!

src/ServerRequestFactory.php
roxblnfk: > PSR implementations must add Host and make it the first one.
> https://tools.ietf.org/html/rfc7230#section-5.4

This requirement applies ONLY to the text format of the HTTP request. In object representation (PSR implementations), it all depends on the implementation of the MessageInterface.

You should follow this:
https://www.php-fig.org/psr/psr-7/#host-header

src/ServerRequestFactory.php
roxblnfk: > Yes.

And why is this case incorrect?
https://tools.ietf.org/html/rfc7230#section-5.4
In some cases the Host header can be empty but this header is required

src/ServerRequestFactory.php
roxblnfk: Why add empty values when even the header for them was not sent by the client?

src/ServerRequestFactory.php
devanych: In this case, the title will be set if it is not there!

This is also necessary when emiting headers.

src/ServerRequestFactory.php
devanych: I agree, need to add a condition for the host here.

src/ServerRequestFactory.php
devanych: This will not be converted to headers. The following parameters will be converted to headers: `REDIRECT_HTTP_*`, etc. Look carefully at the code!

src/ServerRequestFactory.php
devanych: Empty values will not be added. Look carefully at the code!

src/ServerRequestFactory.php
roxblnfk: Can you write tests for this case?

devanych: Yes

src/ServerRequestFactory.php
roxblnfk: or.. for all headers?
https://tools.ietf.org/html/rfc7230#section-3.2
```
header-field   = field-name ":" OWS field-value OWS
field-value    = *( field-content / obs-fold )
```

src/ServerRequestFactory.php
roxblnfk: https://github.com/yiisoft/yii-web/pull/357/files#r584582911

